PR #8, opened 2022-09-22
Install jest & cypress globally as dev-dependency

saneDG:
- Installed Jest globally for both packages
- Installed Cypress globally (previously installed in frontend only)
  - Moved config and example test from frontend to project root


mehtis (approved):

NinaDang97 (approved):

NinaDang97:
@saneDG maybe you can update Readme.doc how to run Cypress (e.g. what script command)?

PR #9, opened 2022-09-22
Update README.md

saneDG:
- updated readme to include steps to run e2e tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,4 +79,27 @@ npm run dev
 
 ### Testing
 
+> ❕ We are in process of building automated testing/CI-pipeline. For now all tests needs to be run manually.
+
+Running e2e/component tests with Cypress.
+
+1. Start local server
+```
+npm run dev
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,4 +79,27 @@ npm run dev
...
+
+1. Start local server
+```
+npm run dev
+```
+
+2. Open Cypress GUI
+
+```
+npx run cypress
```
NinaDang97, line 94: For some reason, this command doesn't work for me. However, `npx cypress open` does work that opens GUI app
Screenshot 2022-09-22 at 13 49 04

saneDG: Typed wrong command here. My bad. Correct command should be in fact `npx cypress open`

saneDG, line 94:
```suggestion
npx cypress open
```

NinaDang97 (approved):

PR #17, opened 2022-09-30
Feature/rtenu-53 accessibility tools

saneDG:
- Add [eslint-plugin-jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y)
- Configure accessibility settings. Set as "recommended".

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,6 @@
 {
-  "extends": ["prettier"],
-  "plugins": ["prettier"],
+  "extends": ["prettier", "plugin:jsx-a11y/strict"],
+  "plugins": ["prettier", "jsx-a11y"],
   "rules": {
     "prettier/prettier": ["error"]
   }
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,5 @@
 {
-  "extends": ["prettier", "plugin:jsx-a11y/strict"],
+  "extends": ["prettier", "plugin:jsx-a11y/recommended"],
   "plugins": ["prettier", "jsx-a11y"],
   "rules": {
     "prettier/prettier": ["error"]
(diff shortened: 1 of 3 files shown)


.eslintrc
```diff
@@ -1,6 +1,6 @@
 {
-  "extends": ["prettier"],
-  "plugins": ["prettier"],
+  "extends": ["prettier", "plugin:jsx-a11y/strict"],
```
NinaDang97, line 2: `"plugin:jsx-a11y/recommended"` would be better and less restrictive for a starting point

mehtis: How strict is strict? Because we're doing software that needs to be accessible, strict rules might be what we should be using.

NinaDang97: `recommended` stills notifies us with similar accessibility errors that `strict` does, except `recommended` gives us options to customize in a few cases (ref: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#rule-strictness-in-different-modes)

saneDG: Idea here was to try strict rules first and we can fallback to recommended if they unnecessarily slow development down. It's easier to go that way than tighten the rules after changes has been applied.

mehtis: Checked a few other repos, and dvk seems to be using recommended. So based on that I'd say that it should be enough 🙂

saneDG, line 2:
```suggestion
  "extends": ["prettier", "plugin:jsx-a11y/recommended"],
```

saneDG: Changed setting to recommended. Thanks for mentioning this @NinaDang97!

mehtis (approved):
You should add the dash to the PR name (e.g. rentu-53 here) so that the PR can be more easily found later on/linking to Jira will work. It's annoying that it has to be done each time, but it'll be of help in the long run.

PR #19, opened 2022-10-03
chore: RTENU-52 Initial version of CDK stack

mehtis:
CDK Pipelines stack and frontend S3 bucket

### DIFF
--- a/lib/config.ts
+++ b/lib/config.ts
@@ -1,6 +1,6 @@
 // Inspiration from https://github.com/finnishtransportagency/hassu/blob/main/deployment/lib/config.ts
 
-export function getEnv(name: string) {
+export function getEnvOrFail(name: string) {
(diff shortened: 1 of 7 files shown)


saneDG (approved):

lib/config.ts
```diff
@@ -0,0 +1,58 @@
+// Inspiration from https://github.com/finnishtransportagency/hassu/blob/main/deployment/lib/config.ts
```
NinaDang97, line 1: Should/can this env config file be used for all packages simultaneously?

NinaDang97 (approved):

lib/config.ts
mehtis: That is the plan. That's what the `getRataExtraStackConfig` is representing. There's just no values there yet.

PR #44, opened 2022-11-04
ci: RTENU-97 Added bastion host for dev

mehtis:
docs: Documented bastion usage


mehtis:
TODO: Update instance id once new instance is up

NinaDang97 (approved):

PR #50, opened 2022-11-09
Feature/rtenu-90 database connection

saneDG:
- Configure local postgres database
- Install and configure Prisma ORM tools
  - [prisma migrate](https://www.prisma.io/migrate)
  - [prisma client](https://www.prisma.io/client)
  - configured to bundle prisma on synth
- Add instructions to run database locally
- Add initial database schema and models
- Two lambdas as example to generate new items to database

TODO: Get DB credentials from AWS Parameter Store and test changes in dev environment.


README.md
```diff
@@ -164,6 +164,61 @@ and then you can connect to bastion host using AWS SSM on port 3002. These are t
...
+Get your IP address
+
+```
+ipconfig getifaddr en0
+```
+
+Copy address to packages/server/.env file
+
+```
+DATABASE_URL="postgresql://root:root@192.168.1.123:5432/test_db?schema=public"
```
saneDG, line 190:
```suggestion
DATABASE_URL="postgresql://root:root@<IP_ADDRESS>:5432/test_db?schema=public"
```

packages/server/prisma/migrations/20221108134009_init/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- CreateTable
+CREATE TABLE "Test" (
```
mehtis, line 2: Is this Test-table something we want to merge to main?

mehtis (changes requested):

mehtis:
Not everything I commented on is needed to be changed. Only the .gitignore is a definite requirement, with others requiring changes only if need be.

docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.8'
+
+services:
+  db:
+    container_name: pg_container
+    image: postgres
+    restart: always
+    environment:
```
NinaDang97, line 8: Should these db's env variables be public here?

lib/rataextra-backend.ts
```diff
@@ -74,8 +104,22 @@ export class RataExtraBackendStack extends NestedStack {
...
+
+    const listUsers = this.createNodejsLambda({
+      ...prismaParameters,
+      name: 'list-users',
+      relativePath: '../packages/server/lambdas/list-users.ts',
+    });
+
     // Add all lambdas here to add as alb targets
     const lambdas: ListenerTargetLambdas[] = [
+      { lambda: listUsers, priority: 70, path: ['/api/users'] },
```
NinaDang97, line 121: @mehtis is it a good idea we number every lambda request's priority manually ourselves? Could a duplicate or wrong priority number happen if the request number grows later? (just curious :octocat: )

packages/server/prisma/schema.prisma
```diff
@@ -0,0 +1,42 @@
...
+model Profile {
+  id     Int     @id @default(autoincrement())
+  bio    String?
+  user   User    @relation(fields: [userId], references: [id])
+  userId Int     @unique
+}
+
+model User {
+  id      Int      @id @default(autoincrement())
+  email   String   @unique
```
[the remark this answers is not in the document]
NinaDang97: we may also need an attribute `isFirstLogin: boolean` for landing page ❓

lib/rataextra-backend.ts
mehtis: I'd expect the synth to fail, if there were duplicate numbers, so that wouldn't go through. My best guess, though. It is possible that a priority number could be in the wrong order. But this is generally an issue only with wildcard values. If we generally use specific paths, then the order doesn't really matter. Maybe a comment could be added here noting to keep the list ordered by priority and to be extra careful with wildcard paths @saneDG

packages/server/prisma/schema.prisma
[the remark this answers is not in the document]
saneDG: These models are here only for example. We ca add/remove these however we like in the future.

packages/server/prisma/migrations/20221108134009_init/migration.sql
saneDG: Yeah this model is unnecessary

packages/server/.gitignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+# Keep environment variables out of version control
+.env
```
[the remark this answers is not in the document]
saneDG: Prisma might have generated this file for some reason. .gitignore in root should indeed be enough.

packages/server/prisma/schema.prisma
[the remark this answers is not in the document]
mehtis: Having these example models will already cause migrations in the database (even when setting up the production database). Not sure if is a good thing... 🤔 Maybe we should already look into designing some first real tables and use those here instead of examples.

docker-compose.yml
saneDG: This should not be problem

packages/server/prisma/schema.prisma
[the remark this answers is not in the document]
mehtis: Maybe even just a few fields for a first table at this point, just to get some real fields for now, but without having to design the whole thing blocking this merge.

packages/server/lambdas/database-client/index.ts
```diff
@@ -0,0 +1,24 @@
...
+    })
+    .catch((err) => {
+      console.log('error: ', err);
+    });
+};
+
+process.env.DATABASE_URL = `postgresql://${getParameter(process.env.SSM_DATABASE_NAME_ID)}:${encodeURIComponent(
+  getParameter(process.env.SSM_DATABASE_PASSWORD_ID) ?? '',
+)}@${getParameter(process.env.SSM_DATABASE_DOMAIN_ID)}:5432/${getParameter(
+  process.env.SSM_DATABASE_NAME_ID,
```
NinaDang97, line 21: Are these SSM_DATABASE_xxx variables already declared in config.ts file?

lib/rataextra-backend.ts
saneDG: Added comment to guide priority usage

lib/rataextra-backend.ts
```diff
@@ -62,6 +64,54 @@ export class RataExtraBackendStack extends NestedStack {
...
+    });
+
+    const ssmDescribeParametersPolicy = new PolicyStatement({
+      actions: ['ssm:DescribeParameters'],
+      resources: ['*'],
+    });
+
+    const ksmDecryptPolicy = new PolicyStatement({
+      actions: ['kms:Decrypt'],
+      resources: ['arn:aws:kms:eu-west-1:178238255639:key/6cd436ad-f1f8-479f-aa56-da5a3f7a0711'],
```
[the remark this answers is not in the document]
mehtis: Later, this might be something to be defined higher in the stack and then passed down, if other resources need access to this.
👍 for specifying instead of using *!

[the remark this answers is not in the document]
mehtis: "These values" include the key being decrypted, as mentioned in the first comment.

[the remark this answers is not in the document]
mehtis: Instead of using a Policy, you could import the key with alias (something like `const ssmDecryptKeyAlias = kms.Alias.fromAliasName(this, 'myKey', 'alias/aws/ssm')`) and then use `ssmDecryptKeyAlias.grantDecrypt(lambda)` to grant decryption rights.

mehtis (approved):

PR #58, opened 2022-11-21
Feat/RTENU-110 search filter view

NinaDang97:


packages/frontend/src/components/Search/FilterSearchData.tsx
```diff
@@ -0,0 +1,121 @@
...
+  CHECKBOX = 'checkbox',
+}
+
+export interface IItem {
+  name: string;
+  type?: ItemTypeEnum;
+  items?: IItem[];
+}
+
+export interface IFilterSearchData {
```
mehtis, line 14: This and `IItem` have the same contents. Do they need to be separate? Can one maybe extend other?

packages/frontend/src/components/Search/FilterSearchData.tsx
```diff
@@ -0,0 +1,121 @@
...
+        name: 'PowerPoint',
+      },
+      {
+        type: ItemTypeEnum.CHECKBOX,
+        name: 'Word',
+      },
+    ],
+  },
+  {
+    name: 'Aika',
```
mehtis, line 49: These are definitely something that need to be created dynamically up to the current year. Some helper function that gets (currentYear, 1980) as parameters?

packages/frontend/src/assets/data/aineistoluokka.json
```diff
@@ -0,0 +1,46 @@
+[
```
mehtis, line 1: Having a filename be in Finnish is a bit weird. But it is under assets, so it might be okay. Leaving this one to percolate for now 🤔

NinaDang97: How about material class or category class? Does it tell full meaning of "aineistoluokka"?

packages/frontend/src/components/Search/FilterSearchData.tsx
NinaDang97: Do you mean we take start year as of 1980 and end year as this current year, and divide by 10 years which result in 5 group of years? (Currently have only helper `generateYearsBetween` that intakes start and end year :/ which is indeed more hard-coding)

packages/frontend/src/assets/data/aineistoluokka.json
mehtis: I can't really say for sure, as I'm no domain expert. But maybe material class would be fine. Too bad `.json` can't have comments...

packages/frontend/src/components/Search/FilterSearchData.tsx
mehtis: Yeah. One option that'll last for 7 years is to have `'2020-2022'` be replaced in a way that the end date is current year. But come 2030, this solution would no longer hold. Someone remembering to do this during the next 7 years if not done now does sounds a bit unlikely as well 😄

mehtis (approved):

PR #68, opened 2022-11-28
Feat/RTENU-108 search result page

NinaDang97:
**Done:**
- Search result page
- Use mock data for node list
- Filter tags

**To-do:**
- Filter search's checkbox indeterminate status (year range checkbox)
- Should re-arrange filter search data before sending to server


packages/frontend/src/contexts/SearchContext.tsx
```diff
@@ -3,17 +3,35 @@ import React, { useState } from 'react';
 export const SearchContext = React.createContext({
   query: '',
   queryHandler: (_: string) => {},
+  checkedList: [''],
+  checkedListHandler: (itemName: string) => {},
 });
 
 export const SearchContextProvider = (props: any) => {
   const [query, setQuery] = useState('');
+  const [checkedList, setcheckedList] = useState<string[]>([]);
```
NinaDang97, line 12: this data type (string[]) is temporary. No information about category name, will be changed in next PR

package.json
```diff
@@ -37,15 +37,18 @@
   "homepage": "https://github.com/finnishtransportagency/ratatiedot-extranet#readme",
   "dependencies": {
     "aws-cdk-lib": "^2.45.0",
+    "axios": "^1.1.3",
     "concurrently": "^7.4.0",
-    "constructs": "^10.0.0"
+    "constructs": "^10.0.0",
+    "moment": "^2.29.4",
```
mehtis, line 43: These days, it would be better not to adopt `moment`, as seen here: https://momentjs.com/docs/ I don't have experience with any of the alternatives, but now's a good chance to gain some 😉

packages/frontend/src/pages/Search/SearchResult.tsx
```diff
@@ -0,0 +1,26 @@
+import { Typography } from '@mui/material';
+import { useSearchParams } from 'react-router-dom';
+import { Tags } from '../../components/Tags';
+import { ContainerWrapper } from '../Landing/index.styles';
+
+// TODO: Use real data fetched from Backend (currently use mock data)
+import mockData from './mockData.json';
```
mehtis, line 7: For future: Using env variable (`ENV=local` or such) to use mockData locally?

packages/frontend/src/pages/Search/NodeItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+      sx={{ paddingBottom: '18px', backgroundColor: row % 2 ? Colors.lightgrey : Colors.white }}
+    >
+      <Grid item mobile={1} tablet={0.5} desktop={0.5}>
+        <Box component="img" src={matchMimeType(content.mimeType)} alt="Logo" />
+      </Grid>
+      <Grid item mobile={11} tablet={11.5} desktop={11.5}>
+        <Typography variant="body1">{name}</Typography>
+        <div style={{ display: 'flex', color: Colors.darkgrey, paddingBottom: '18px' }}>
+          <Typography variant="body1" sx={{ marginRight: '8px' }}>
+            {moment(modifiedAt).format('DD.MM.YYYY')}
```
mehtis, line 49: With moment or whatever we'll use, maybe set the format in a single location and import it from there to more easily maintain consistency.

packages/frontend/src/pages/Search/SearchResult.tsx
NinaDang97: Good question. Should also dev env use mockData if we deploy (ofc for temporary use to display for customer)?

mehtis: In dev we have the real API available, so we should use that. Otherwise we wouldn't be able to test API calls without going to prod 😅

saneDG (approved):
Build was succesfull and I tested manually some features. Looking good 👍

NinaDang97:
> Build was succesfull and I tested manually some features. Looking good 👍

Thanks!!

PR #70, opened 2022-11-28
Feature/rtenu 90 queries

saneDG:
- Use ES modules to make top-level await possible.
- Required changes to make prisma database querying working.
  - Increase lambda timeout value from 5 to 15 seconds


packages/server/lambdas/create-user.ts
```diff
@@ -36,13 +37,10 @@ export async function handleRequest(_event: APIGatewayEvent, _context: Context)
         body: JSON.stringify(res),
         isBase64Encoded: false,
       };
+      console.log('res: ', response.body);
```
mehtis, line 40: Console.log: For testing (mark with TODO, maybe?) or forgotten?

packages/server/lambdas/list-users.ts
```diff
@@ -29,13 +25,10 @@ export async function handleRequest(_event: APIGatewayEvent, _context: Context)
         body: JSON.stringify(res),
         isBase64Encoded: false,
       };
+      console.log('USERS: ', response.body);
```
mehtis, line 28: Same comment about console.log

lib/rataextra-backend.ts
```diff
@@ -222,7 +268,7 @@ export class RataExtraBackendStack extends NestedStack {
     vpc,
     securityGroups,
     memorySize = 1024,
-    timeout = Duration.seconds(5),
+    timeout = Duration.seconds(10),
```
mehtis, line 271: If 5 seconds was too tight, maybe put this to 15 seconds just in case?

lib/rataextra-backend.ts
```diff
@@ -87,6 +88,42 @@ export class RataExtraBackendStack extends NestedStack {
       securityGroups: securityGroups,
     };
 
+    /**
+     * @param {string} projectRoot
+     * @returns {import('esbuild').Plugin}
+     */
+    const createFixDirnamesPlugin = (projectRoot: string) => {
```
mehtis, line 95: This and ESM_REQUIRE_SHIM are a bit much to add in the middle of this infra code. I'd suggest moving these elsewhere and importing them from there.

lib/config.ts
```diff
@@ -37,6 +37,7 @@ export const SSM_DATABASE_NAME = 'rataextra-database-name';
 export const SSM_DATABASE_PASSWORD = 'rataextra-rdspg13-rataextradev-password';
 const SSM_CLOUDFRONT_SIGNER_PUBLIC_KEY = 'rataextra-cloudfront-signer-public-key';
 export const SSM_CLOUDFRONT_SIGNER_PRIVATE_KEY = 'rataextra-cloudfront-signer-private-key';
+export const ESM_REQUIRE_SHIM = `await(async()=>{let{dirname:e}=await import("path"),{fileURLToPath:i}=await import("url");if(typeof globalThis.__filename>"u"&&(globalThis.__filename=i(import.meta.url)),typeof globalThis.__dirname>"u"&&(globalThis.__dirname=e(globalThis.__filename)),typeof globalThis.require>"u"){let{default:a}=await import("module");globalThis.require=a.createRequire(import.meta.url)}})();`;
```
saneDG, line 40: Shim will be exported from here. Ok? @mehtis

mehtis: File is fine, but have it separate from these SSM values. Don't want to mix them up.

mehtis (approved):

PR #75, opened 2022-11-30
Revert "Feat: RTENU-87 Check frontend authorisation with lambda@edge"

mehtis:
Reverts finnishtransportagency/ratatiedot-extranet#73
Lambda@edge requires setting specific account at synth, but adding it now would require removing the current stack.


mehtis:
validateUser is something I'll remove again, but revering it just came as a package deal here.

PR #90, opened 2022-12-08
Feat/RTENU-83 Alfresco API integration

NinaDang97:


packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!alfrescoAPIKey) {
+      alfrescoAPIKey = await getSecuredStringParamter(alfrescoAPIKeyName);
+    }
+    if (!alfrescoAPIUrl) {
+      alfrescoAPIUrl = await getSecuredStringParamter(alfrescoAPIUrlName);
+    }
+    const options = {
+      headers: {
+        'X-API-Key': alfrescoAPIKey,
+        // Cookie: '',
```
NinaDang97, line 41: Hi, do I need to get cookie from `event.headers` or `cloudFrontSigner.getSignedCookie` and pass it here?

lib/config.ts
```diff
@@ -58,6 +60,8 @@ export const getRataExtraStackConfig = (scope: Construct) => ({
   dmzApiEndpoint: getSSMStringParameter(scope, SSM_DMZ_API_DOMAIN_NAME),
   databaseDomain: getSSMStringParameter(scope, SSM_DATABASE_DOMAIN),
   jwtTokenIssuer: getSSMStringParameter(scope, SSM_JWT_TOKEN_ISSUER),
+  alfrescoAPIKeyName: getSSMStringParameter(scope, SSM_ALFRESCO_API_KEY_NAME),
```
mehtis, line 63: This will try to get the value. Unsuccessfully, as this is a SecuredString. Instead, the current solution is to pass the e.g. `SSM_ALFRESCO_API_KEY_NAME` directly as a string to the specified stack (rataextra-backend.ts in this case). I'm not happy with the current situation and might refactor it, but that's the current situation.

packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,72 @@
...
+import axios from 'axios';
+
+import { getRataExtraLambdaError } from '../../utils/errors';
+import { log } from '../../utils/logger';
+import { getUser, validateReadUser } from '../../utils/userService';
+
+let alfrescoAPIKey: string | null = null;
+let alfrescoAPIUrl: string | null = null;
+const alfrescoAPIKeyName = process.env.ALFRESCO_API_KEY_NAME || '';
+const alfrescoAPIUrlName = process.env.ALFRESCO_API_URL_NAME || '';
```
mehtis, line 12: URL you can get the value from CDK (which you actually already do). So need to call it name.

packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const getSecuredStringParamter = async (name: string) => {
```
mehtis, line 14: Move this to utils or something and import from there. We can then do that to the existing database stuff as well. No need to duplicate code.

packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+};
+
+const searchByTerm = async (term: string) => {
+  try {
+    if (!alfrescoAPIKey) {
+      alfrescoAPIKey = await getSecuredStringParamter(alfrescoAPIKeyName);
+    }
+    if (!alfrescoAPIUrl) {
+      alfrescoAPIUrl = await getSecuredStringParamter(alfrescoAPIUrlName);
```
mehtis, line 36: Like I mentioned, you should have URL already from process.env, as it's not a secured string.

lib/config.ts
```diff
@@ -35,6 +35,8 @@ const SSM_JWT_TOKEN_ISSUER = 'rataextra-jwt-token-issuer';
 export const SSM_DATABASE_DOMAIN = 'rataextra-database-domain';
 export const SSM_DATABASE_NAME = 'rataextra-database-name';
 export const SSM_DATABASE_PASSWORD = 'rataextra-rdspg13-rataextradev-password';
+export const SSM_ALFRESCO_API_KEY_NAME = 'rataextra-alfresco-api-key';
+export const SSM_ALFRESCO_API_URL_NAME = 'rataextra-alfresco-api-url';
```
mehtis, line 39: I'd drop the _NAME. from these, as it's not descriptive.

lib/config.ts
```diff
@@ -58,6 +60,8 @@ export const getRataExtraStackConfig = (scope: Construct) => ({
   dmzApiEndpoint: getSSMStringParameter(scope, SSM_DMZ_API_DOMAIN_NAME),
   databaseDomain: getSSMStringParameter(scope, SSM_DATABASE_DOMAIN),
   jwtTokenIssuer: getSSMStringParameter(scope, SSM_JWT_TOKEN_ISSUER),
+  alfrescoAPIKeyName: getSSMStringParameter(scope, SSM_ALFRESCO_API_KEY_NAME),
+  alfrescoAPIUrlName: getSSMStringParameter(scope, SSM_ALFRESCO_API_URL_NAME),
```
mehtis, line 64: This will pass the actual value. Which is good! But as such, drop the Name from the variable name.

packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,72 @@
+import { ALBEvent, Context } from 'aws-lambda';
+import AWS from 'aws-sdk';
+import axios from 'axios';
+
+import { getRataExtraLambdaError } from '../../utils/errors';
+import { log } from '../../utils/logger';
+import { getUser, validateReadUser } from '../../utils/userService';
+
+let alfrescoAPIKey: string | null = null;
```
mehtis, line 9: Yay, cache!

mehtis (changes requested):

packages/server/lambdas/alfresco/search.ts
mehtis: Looks good now. Just a small note now: a missing e from Parameter.

event.json
```diff
@@ -0,0 +1,6 @@
+{
+  "headers": {
+    "cookie": ""
```
NinaDang97, line 3: Pasting your own cookie here. Run invoke command with `-- -e event.json`

NinaDang97: Waiting for cookie removal...

packages/server/lambdas/alfresco/search.ts
```diff
@@ -12,17 +12,19 @@ let alfrescoAPIUrl: string | null = null;
...
     // ALBEvent's body type is string | null
+    log.info(body, 'POST body request');
     const parsedBody = body ? JSON.parse(body) : {};
+    log.info(parsedBody, 'Body parsing...');
     const bodyRequest = searchQueryBuilder({
       searchParameters: parsedBody.searchParameters,
       page: parsedBody.page,
       language: parsedBody.language,
     });
+    log.info(bodyRequest, 'Complete body request');
```
mehtis, line 27: TODO: Add support for setting log-level to debug and change this to debug.

mehtis: So not a problem to be fixed here, but do add the TODO-marking

packages/server/lambdas/alfresco/search.ts
```diff
@@ -0,0 +1,56 @@
...
+const alfrescoAPIKeyName = process.env.ALFRESCO_API_KEY || '';
+const alfrescoAPIUrlName = process.env.ALFRESCO_API_URL || '';
+
+const searchByTerm = async (term: string) => {
+  try {
+    if (!alfrescoAPIKey) {
+      alfrescoAPIKey = await getSecuredStringParamter(alfrescoAPIKeyName);
+    }
+    if (!alfrescoAPIUrl) {
+      alfrescoAPIUrl = await getParamter(alfrescoAPIUrlName);
```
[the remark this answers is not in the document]
mehtis: Still a valid comment. Ask me for help if needed.

lib/rataextra-backend.ts
```diff
@@ -183,13 +187,26 @@ export class RataExtraBackendStack extends NestedStack {
       relativePath: '../packages/server/lambdas/return-login.ts',
     });
 
+    const alfrescoSearch = this.createNodejsLambda({
+      ...genericLambdaParameters,
+      environment: {
```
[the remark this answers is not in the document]
mehtis: I'm adding this in my PR.

packages/server/lambdas/database-client/index.ts
```diff
@@ -1,37 +1,23 @@
...
-    const [databaseName, databaseDomain, databasePassword] = await Promise.all([
-      getParameter(process.env.SSM_DATABASE_NAME_ID),
-      getParameter(process.env.SSM_DATABASE_DOMAIN_ID),
-      getSecureStringParameter(process.env.SSM_DATABASE_PASSWORD_ID),
-    ]);
+    if (!databaseName || !databaseDomain || !databasePassword) {
+      [databaseName, databaseDomain, databasePassword] = await Promise.all([
+        getParameter(databaseNameId),
+        getParameter(databaseDomainId),
+        getParameter(databasePasswordId),
```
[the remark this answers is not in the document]
mehtis: Still valid comment.

mehtis (approved):

PR #93, opened 2022-12-09
Feature/rtenu 42 localization

saneDG:
- Install i18next, i18next-browser-languagedetector & react-i18next libraries
- Add translation files
  - Use translations in Landing page as an example

### DIFF
--- a/packages/frontend/package.json
+++ b/packages/frontend/package.json
@@ -12,7 +12,6 @@
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
     "i18next": "^22.1.4",
-    "i18next-browser-languagedetector": "^7.0.1",
     "react": "^18.2.0",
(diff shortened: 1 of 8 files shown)


packages/frontend/package.json
```diff
@@ -11,8 +11,11 @@
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@testing-library/user-event": "^13.5.0",
+    "i18next": "^22.1.4",
+    "i18next-browser-languagedetector": "^7.0.1",
```
mehtis, line 15: YAGNI? This is something we won't be using right now, so I'd say that we should leave this out (maybe add a comment where you could use this) as it's just extra cruft at this point.

mehtis (changes requested):

mehtis (approved):

PR #94, opened 2022-12-12
Fix: parse page title into route name properly

NinaDang97:


packages/frontend/src/components/NavBar/MenuItems.tsx
```diff
@@ -20,6 +20,14 @@ export interface IMenuItem {
   children?: any;
 }
 
+const hasWhitespace = (name: string) => {
+  return /\s/.test(name);
+};
+
+export const getRouterName = (name: string) => {
+  return hasWhitespace(name) ? name.split(' ').join('-').toLocaleLowerCase() : name.toLocaleLowerCase();
```
mehtis, line 28: Could this be simplified to using `.replace`? Could also skip the ternary and just do that, because you would/could be using the same regex there. And to top this off, as this is a single line of code, this could skip the {}.

saneDG (approved):
Approved after @mehtis comment has been resolved

mehtis (approved):

PR #96, opened 2022-12-13
Bugfix/postgresql lambda

saneDG:
Fix two lambdas by nest database queries in try-catch block.


packages/server/lambdas/list-users.ts
```diff
@@ -9,26 +9,19 @@ const database = await DatabaseClient.build();
 export async function handleRequest(event: ALBEvent, _context: Context) {
   try {
     const user = await getUser(event);
-    log.info(user, 'Listing users');
```
mehtis, line 12: Return this logging (so it won't be forgotten that easily when real things start to happen.

mehtis (changes requested):

packages/server/lambdas/create-user.ts
```diff
@@ -11,12 +11,7 @@ export async function handleRequest(event: ALBEvent, _context: Context) {
     const user = await getUser(event);
     log.info(user, 'Creating temp users');
     await validateWriteUser(user);
-  } catch (err) {
-    log.error(err);
-    return getRataExtraLambdaError(err);
-  }
-  await database.user
-    .create({
+    const cretedUser = await database.user.create({
```
NinaDang97, line 14: typo `createdUser`?

mehtis (approved):

PR #97, opened 2022-12-14
Feat: RTENU-81 Category data database

mehtis:
* API endpoints for getting custom fields in a page (/api/database/get-page-contents) and for getting files on a page (/api/alfresco/list-files)
* Database fields for linking frontend pages to alfresco
* Changed some scripts in package.json to use local in naming
* Refactored some existing CDK code to use more shared values (esp. for ENV)
* Refactored some backend code to get static, non-secret values from ENV instead of Parameter Store
* Added script for populating local db
* Added new fields to search queries
* Feat and local environments now set logging level to debug


packages/server/lambdas/alfresco/searchQueryBuilder/types.ts
```diff
@@ -32,6 +33,10 @@ export type Paging = {
   skipCount: number;
 };
 
+export enum Include {
```
NinaDang97, line 36: "Include" name is a bit confusing. At least I don't understand what its purpose based on the name?

mehtis: That's what the field is called in the API, but it definitely is not descriptive. I'll think of something better. Thanks!

mehtis: Updated to AdditionalFields.

packages/server/prisma/schema.prisma
```diff
@@ -11,6 +11,7 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
+// Test models
```
saneDG, line 14: We should remove test models after this and start testing api/queries with "Real" models. This should be done in another PR where we also remove those deprecated lambdas and endpoints.

mehtis: Yup. Once we have tested this as working in main, we can start removing those.

packages/server/populate-local-db.sh
```diff
@@ -0,0 +1,11 @@
...
+set -a; source .env; set +a
+psql -h $LOCAL_IP -p $DB_PORT -d $DATABASE_NAME -U $DATABASE_USER << END_OF_SCRIPT
+TRUNCATE "CategoryDataContents", "CategoryDataBase" restart identity;
+
+INSERT INTO "CategoryDataBase" ("rataextraRequestPage", "alfrescoFolder", "writeRights")
+VALUES ('/linjakaaviot', '/123', 'Rataextra_kirjoitus_linjakaaviot');
+
+INSERT INTO "CategoryDataContents" ("baseId", "fields")
+VALUES (1, '{"test": "test"}')
+END_OF_SCRIPT
```
saneDG, line 11: We could seed our database using TS [also](https://www.prisma.io/docs/guides/database/seed-database#example-seed-scripts).

But this approach works too 👍🏻

mehtis: Definitely something to improve upon. And then there's the matter of where to store the dev-env values for population. Not sure if we want to have those in public like this, if they'd be populated from here.

saneDG (approved):

packages/server/lambdas/alfresco/searchQueryBuilder/__tests__/luceneQueryBuilder.test.ts
```diff
@@ -126,23 +126,37 @@ describe('Lucene Query Builder', () => {
...
-  });
-  describe('lucenePagination', () => {
     it('should return given positive pagination page', () => {
       expect(luceneQueryBuilder.pagination(5)).toEqual({ maxItems: 10, skipCount: 5 });
     });
-  });
-
-  describe('lucenePagination', () => {
     it('should return zero given negative pagination page', () => {
       expect(luceneQueryBuilder.pagination(-4)).toEqual({ maxItems: 10, skipCount: 0 });
```
NinaDang97, line 147: Just a minor fix, could you change all `maxItems` to 25 in this test file? My previous PR increased the limit

packages/server/lambdas/alfresco/search.ts
```diff
@@ -1,48 +1,30 @@
...
     const bodyRequest = searchQueryBuilder({
       searchParameters: parsedBody.searchParameters,
       page: parsedBody.page,
       language: parsedBody.language,
     });
-    log.info(bodyRequest, 'Complete body request');
+    log.debug(bodyRequest, 'Complete body request');
+
+    const alfrescoAPIUrl = getAlfrescoUrlBase();
+    const options = await getAlfrescoOptions(uid, { 'Content-Type': 'application/json;charset=UTF-8' });
```
NinaDang97, line 23: Thanks for refactoring!

packages/server/lambdas/alfresco/searchQueryBuilder/__tests__/index.test.ts
```diff
@@ -0,0 +1,17 @@
...
+
+describe('searchQueryBuilder', () => {
+  it("don't have additionalFields alongside query if not given", () => {
+    expect(searchQueryBuilder({ language: QueryLanguage.LUCENE, searchParameters: [], page: 0 })).toEqual({
+      query: {
+        language: 'lucene',
+        query: '',
+      },
+      paging: {
+        maxItems: 10,
```
NinaDang97, line 12: It's 25 now :)

packages/server/lambdas/alfresco/list-files.ts
```diff
@@ -0,0 +1,79 @@
...
+    return response.data;
+  } catch (err) {
+    throw err;
+  }
+};
+
+const database = await DatabaseClient.build();
+
+let fileEndpointsCache: Array<CategoryDataBase> = [];
+
```
NinaDang97, line 44: Should we document by commenting in each `handleRequest` function what endpoint, method, body request structure,...? It'll be easier for team members to test

mehtis: This has been done now. Please share your thoughts!

PR #101, opened 2022-12-20
fix: Lambda building when not doing local stack

mehtis:
Might be a temporary fix, but this should get the Prisma stuff working for now.
When deploying feat-stacks, it'll do it the slow way, but at least stacks with local env use bundling.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -23,6 +23,12 @@ export const isProductionStack = (stackId: string, rataExtraEnv: RataExtraEnviro
  */
 export const isFeatOrLocalStack = (rataExtraEnv: RataExtraEnvironment) =>
   rataExtraEnv === ENVIRONMENTS.feat || rataExtraEnv === ENVIRONMENTS.local;
+
+/**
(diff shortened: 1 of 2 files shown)


lib/utils.ts
```diff
@@ -23,6 +23,12 @@ export const isProductionStack = (stackId: string, rataExtraEnv: RataExtraEnviro
  */
 export const isFeatOrLocalStack = (rataExtraEnv: RataExtraEnvironment) =>
   rataExtraEnv === ENVIRONMENTS.feat || rataExtraEnv === ENVIRONMENTS.local;
```
NinaDang97, line 25: `isFeatOrLocalStack` to `isFeatStack`, no need to combine?

NinaDang97 (approved):

lib/utils.ts
mehtis: We check for local, because all other environments are to be deploy to the cloud, and then the DB calls will fail because of the Prisma problem. This way we can use esbuild at least when doing LOCAL-stuff.

PR #102, opened 2022-12-20
Feat/RTENU-109 alfresco search view

NinaDang97:
**Done**
- Adopt year picker (from year - to year)
- "Päivitä tulokset" will update the search result and filtered tags 
- Search page will use mock data if fail to fetch API
- Remove Home page. Etusivu page uses '/' route

**TODO**
- "Poista kaikki suodattimet" button doesn't reset checkboxes currently, only year picker 
- No filter data is used


packages/frontend/src/pages/Search/SearchResult.tsx
```diff
@@ -9,17 +14,50 @@ import { NodeItem } from './NodeItem';
...
+      };
+      const response = await axios.post('/api/alfresco/search', body);
+      return response.data;
+    },
+    onSuccess: (res) => {
+      setData(res);
+    },
+    // temporary
+    // TODO:
+    onError: (err) => {
```
NinaDang97, line 39: This is supposed to show error message in toaster

packages/frontend/src/index.tsx
```diff
@@ -10,7 +10,17 @@ import { theme } from './styles/createTheme';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import './i18n';
 
-const queryClient = new QueryClient();
+const queryClient = new QueryClient({
```
mehtis, line 13: Once we start adding more queries, these should be moved to a centralised location instead of in the middle of other code. But that can be done later.

packages/frontend/src/pages/Search/SearchResult.tsx
```diff
@@ -9,17 +14,51 @@ import { NodeItem } from './NodeItem';
 
 export const SearchResult = () => {
   const [searchParams] = useSearchParams();
+  const [data, setData] = useState<any>({}); // TODO: shouldn't be `any` type
   const query = searchParams.get('query');
 
+  // TODO: Move this to separate api folder?
```
mehtis, line 20: You were ahead of me!

packages/frontend/src/components/Search/FilterSearch.tsx
```diff
@@ -1,52 +1,51 @@
...
+import { EMimeType } from '../../constants/Data';
 
 type FilterSearchProps = {
   openFilter: boolean;
   toggleFilter: any;
 };
 
-const FilterSearchItem = (props: IItem) => {
+interface IFilterSearchItem extends IItem {
+  checkboxes: any;
```
mehtis, line 27: Any way to do _some_ typing here?

saneDG (approved):

mehtis (approved):
Non-blocking comments.

packages/frontend/src/components/Search/FilterSearch.tsx
NinaDang97: I'll change those `any` to proper types after this PR, include other of your comments too

PR #108, opened 2022-12-20
Feat/RTENU-109 connect frontend to alfresco api

NinaDang97:
- Aim to trigger useQuery and refresh component after submit enter in search bar
- Have some basic error alert and loading


mehtis (approved):
Small question with the `data`, but once that is clear, it's okay to merge this.

PR #109, opened 2022-12-21
clear cookies before logout

saneDG:
"Return" cookie keeps already authenticated user out of login form. This causes problem when unauthenticated user tries navigate login form but we keep sending user back to frontend app.

- Remove "Return" cookie before logout invocation.

### DIFF
--- a/packages/frontend/src/routes.tsx
+++ b/packages/frontend/src/routes.tsx
@@ -1,4 +1,4 @@
-import { createBrowserRouter, RouteObject } from 'react-router-dom';
+import { createBrowserRouter, RouteObject, redirect } from 'react-router-dom';
 
 import { Landing } from './pages/Landing';
 import { Routes } from './constants/Routes';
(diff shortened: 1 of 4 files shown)


NinaDang97 (approved):
Need @mehtis approval too to be sure

packages/frontend/src/routes.tsx
```diff
@@ -33,6 +33,19 @@ const routes: RouteObject[] = [
...
   },
+  {
+    path: Routes.LOGOUT,
+    element: <div>LOGGING OUT...</div>,
+    errorElement: <RootBoundary />, // Send user here whenever error is thrown
+    loader: async () => {
+      console.log('Logging out');
+      document.cookie = 'Return=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=/;';
+      // check if cookie exists then redirect to /
+      if (document.cookie.indexOf('Return') === -1) {
```
mehtis, line 44: What if the cookie isn't removed? What then? Can they get stuck here? Should we offer a button here for manual redirect that activates after a few seconds?

mehtis: This comment doesn't necessarily require code changes, just some food for thought.

saneDG: Added comment about this in the code

mehtis (approved):

PR #112, opened 2022-12-22
Fix/RTENU-109 filter search functionality

NinaDang97:
- "Something went wrong" still exists. Turns out there's no null check for `content.sizeInBytes`
- Map mimetype before sending api request

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,9 +41,11 @@
     "concurrently": "^7.4.0",
     "constructs": "^10.0.0",
     "date-fns": "^2.29.3",
+    "lodash": "^4.17.21",
     "pretty-bytes": "^6.0.0"
   },
   "devDependencies": {
+    "@types/lodash": "^4.14.191",
     "@types/node": "^18.7.23",
     "@typescript-eslint/eslint-plugin": "^5.39.0",
     "@typescript-eslint/parser": "^5.39.0",
--- a/packages/frontend/src/pages/Search/SearchResult.tsx
+++ b/packages/frontend/src/pages/Search/SearchResult.tsx
@@ -10,6 +10,7 @@ import { useContext } from 'react';
 import { SearchContext } from '../../contexts/SearchContext';
 import { formatYear } from '../../utils/helpers';
 import { useSearchParams } from 'react-router-dom';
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -41,9 +41,11 @@
     "concurrently": "^7.4.0",
     "constructs": "^10.0.0",
     "date-fns": "^2.29.3",
+    "lodash": "^4.17.21",
     "pretty-bytes": "^6.0.0"
   },
   "devDependencies": {
+    "@types/lodash": "^4.14.191",
```
mehtis, line 48: Lodash was added twice. Presumably not needed in devDependecies?

packages/frontend/src/pages/Search/SearchResult.tsx
```diff
@@ -21,7 +22,7 @@ export const SearchResult = () => {
     term: query,
     from: formatYear(years[0]),
     to: formatYear(years[1]),
-    fileTypes: checkedList.mime,
+    fileTypes: checkedList.mime.map((mimeType: string) => mimeNamesMapping[mimeType as keyof typeof mimeNamesMapping]),
```
mehtis, line 25: Curious: What does this bit do?

mehtis (changes requested):
Lodash things needs to be fixed, other comment is just so that I understand what is happening.

package.json
NinaDang97: Without "@types/lodash", Typescript can't find a declaration file for module 'lodash'. Otherway would be manually declare module lodash in types.d.ts but this faster

mehtis: Reading comprehension fail

mehtis (approved):

package.json
NinaDang97: Or I misunderstood your question

packages/frontend/src/pages/Search/SearchResult.tsx
NinaDang97: checkedList.mime contains array of checkbox names which are slightly different from the mime names

package.json
mehtis: The fail was on me. Didn't notice the @types

PR #120, opened 2022-12-28
Feat/RTENU-129 search history clickable links

NinaDang97:
Screenshot 2022-12-28 at 13 49 33

**Done**
- Improve LocalStorageHelper class functionality: remove duplicates, fix `null` bug (first time search adds null to array)
- Recent search is clickable, add clicked term to local storage


saneDG (approved):
Couple bugs that needs to be fixed. For me these changes look good and fixes can be done in another PR if you want.
- Need to click input element twice to be able to type in it
- Only way to close search history is to click any item.

PR #123, opened 2023-01-02
Feat: RTENU-126 page contents updating

mehtis:
* Added writeRights validation support
* Added db fields update to test writeRights check
* Added static admin role
* Added http methods to alb
* Refactored lambda policy addition to avoid repetition
* Removed unnecessary awaits from validation calls
* Updated sam:invoke instructions, as AWS profile is needed now for some reason
* Added support for using local db with sam
* Updated getRouterNames to remove even more special characters
* Hack fix for api 403 errors redirecting to /index.html, using 401 instead

### DIFF
--- a/packages/server/lambdas/database/client/index.ts
+++ b/packages/server/lambdas/database/client/index.ts
@@ -1,4 +1,5 @@
 import { PrismaClient } from '@prisma/client';
+import { ENVIRONMENTS } from '../../../../../lib/config';
 import { getParameter, getSecuredStringParameter } from '../../../utils/parameterStore';
 
 const environment = process.env.ENVIRONMENT;
@@ -13,7 +14,7 @@ let databasePassword: string | null = null;
 export class DatabaseClient {
   public static async build(): Promise<PrismaClient> {
     let DATABASE_URL: string;
-    if (environment === 'local') {
+    if (environment === ENVIRONMENTS.local) {
       DATABASE_URL = 'postgresql://root:root@host.docker.internal:5432/test_db?schema=public';
     } else {
       if (!databaseName || !databaseDomain || !databasePassword) {
--- a/packages/server/lambdas/database/client/index.ts
+++ b/packages/server/lambdas/database/client/index.ts
@@ -1,6 +1,7 @@
 import { PrismaClient } from '@prisma/client';
 import { getParameter, getSecuredStringParameter } from '../../../utils/parameterStore';
 
+const environment = process.env.ENVIRONMENT;
 const databaseNameId = process.env.SSM_DATABASE_NAME_ID || '';
 const databaseDomainId = process.env.SSM_DATABASE_DOMAIN_ID || '';
 const databasePasswordId = process.env.SSM_DATABASE_PASSWORD_ID || '';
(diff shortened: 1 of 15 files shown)


packages/server/utils/userService.ts
```diff
@@ -82,25 +83,27 @@ export const getUser = async (event: ALBEvent): Promise<RataExtraUser> => {
...
-      throw new RataExtraLambdaError('Forbidden', 403);
-    }
+/**
+ * Checks if the user has necessary role for write access. Also checks for admin rights. Throws 403 Forbidden if not authorised.
+ * @param {RataExtraUser} user User being validated
+ * @param {string} writeRole Role being validated against
+ */
+export const validateWriteUser = (user: RataExtraUser, writeRole: string): void => {
+  if (isAdmin(user)) {
+    return;
```
NinaDang97, line 102: Or ?
```
if (isAdmin(user) || isWriteUser(user, writeRole)) {
    return;
  } else {
log.error(user, 'Forbidden: User is not an authorised write user');
    throw new RataExtraLambdaError('Forbidden', 403);
}
```

packages/server/lambdas/database/client/index.ts
```diff
@@ -11,15 +13,21 @@ let databasePassword: string | null = null;
...
   public static async build(): Promise<PrismaClient> {
-    if (!databaseName || !databaseDomain || !databasePassword) {
-      [databaseName, databaseDomain, databasePassword] = await Promise.all([
-        getParameter(databaseNameId),
-        getParameter(databaseDomainId),
-        getSecuredStringParameter(databasePasswordId),
-      ]);
-    }
+    let DATABASE_URL: string;
+    const setLocalDatabaseConnection = environment === ENVIRONMENTS.local;
```
NinaDang97, line 17: should be `isLocalDatabaseConnection` ?

packages/server/utils/userService.ts
mehtis: Good call. Updated

packages/server/lambdas/database/client/index.ts
mehtis: Wasn't happy with set, but didn't go for is for some reason. Fixed!

packages/server/lambdas/database/client/index.ts
```diff
@@ -11,15 +13,21 @@ let databasePassword: string | null = null;
...
+    if (setLocalDatabaseConnection) {
+      DATABASE_URL = 'postgresql://root:root@host.docker.internal:5432/test_db?schema=public';
+    } else {
```
NinaDang97, line 20: We are not allowed to get ssm keys if locally? 
How about remove if-else for `setLocalDatabaseConnection` and replace with this?
``` 
let DATABASE_URL = setLocalDatabaseConnection ? "postgresql://root:root@host.docker.internal:5432/test_db?schema=public" : "postgresql://${databaseName}:${databasePassword}@${databaseDomain}:5432/${databaseName}?schema=public"
```

mehtis: If we're running locally, there's no need for SSM stuff.

NinaDang97 (approved):

PR #124, opened 2023-01-03
Hotfix/UI bugs

NinaDang97:
**TODO**
- Tapping outside search bar is supposed to close recent search. However, `onBlur` is fired before `onClick` so it changes some components' original behaviors. I don't have any solution now except replacing `onClick` with `onMouseDown`


mehtis:
What is being fixed here?

mehtis (approved):

PR #129, opened 2023-01-04
fix search history bug without timeout

saneDG:
Make sure everything works before merging this. This is one way to solve onClick/onBlur issue without dangerous timeout method.


NinaDang97:
Thanks @saneDG , this is something I need! Was stuck with recent search's links

PR #134, opened 2023-01-09
Replace <div> with <React.Fragment>

NinaDang97:
Fix responsiveness on mobile:
Screenshot 2023-01-05 at 18 22 27


packages/frontend/src/components/NavBar/MenuList.tsx
```diff
@@ -1,6 +1,7 @@
 import { ExpandLess, ExpandMore } from '@mui/icons-material';
 import { ListItem, ListItemButton, ListItemIcon, ListItemText } from '@mui/material';
-import { useContext } from 'react';
+import React, { useContext } from 'react';
```
mehtis, line 3: Maybe import `{ Fragment }`? Not sure if that helps with any bundling optimisation or not.

mehtis (approved):

PR #137, opened 2023-01-11
feat/RTENU-134 perform sorting query

NinaDang97:


packages/frontend/src/hooks/query/Search.ts
```diff
@@ -9,10 +9,15 @@ export type TAlfrescoSearchProps = {
   to?: string | number;
   fileTypes?: string[];
   page?: number;
+  sort?: any[];
```
mehtis, line 12: Any way to type this at all?

mehtis: Pun not intended

NinaDang97: I'll change to suitable type

mehtis (approved):

PR #140, opened 2023-01-17
Alfresco file upload

saneDG:
- Add endpoint for uploading file
  - Lambda function
  - Alfresco integration
- Remove wildcard imports from infra code
- Add build step 'StripAssetsFromAssembly' to reduce the size files created at build phase. Previously prevented stack from building at all.
- Add test for fileRequestBuilder


lib/rataextra-pipeline.ts
```diff
@@ -40,6 +41,37 @@ export class RataExtraPipelineStack extends Stack {
...
+        'zip -r -q -A $ZIP_ARCHIVE *',
+        'aws s3 cp $ZIP_ARCHIVE s3://$S3_PATH',
+        'object_location=${S3_PATH#*/}',
+        'for bucket in $cross_region_replication_buckets; do aws s3 cp $ZIP_ARCHIVE s3://$bucket/$object_location; done',
+      ],
+      rolePolicyStatements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          resources: ['*'],
+          actions: ['s3:*'],
```
mehtis, line 61: Any chance to remove `*` here?

lib/rataextra-pipeline.ts
```diff
@@ -40,6 +41,37 @@ export class RataExtraPipelineStack extends Stack {
...
+        }),
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          resources: ['*'],
```
mehtis, line 65: `*` here as well

packages/frontend/src/pages/Others/ManagementReports/index.tsx
```diff
@@ -1,3 +1,48 @@
...
+    const selectedFiles = files as FileList;
+    setFile(selectedFiles?.[0]);
+  };
+  const upload = () => {
+    if (!file) return;
+
+    FileUploadService.upload(nodeId, file).catch((err) => {
+      if (err.response && err.response.data && err.response.data.message) {
+      } else {
+        console.log('something went wrong!');
```
mehtis, line 18: Proper error handling? Or at least a TODO. Here and with this whole function. Should this be a function that's imported from a utils or something, as it's going to be needed on so many pages?

packages/frontend/src/pages/Others/ManagementReports/index.tsx
```diff
@@ -1,3 +1,48 @@
...
+    <>
+      <TextField type="file" onChange={selectFile} />
+      <TextField
+        value={nodeId}
+        onChange={(event) => {
+          setNodeId(event.target.value);
+        }}
+      />
+      <Button type="submit" fullWidth variant="contained" sx={{ mt: 3, mb: 2 }} onClick={upload}>
+        Upload
```
mehtis, line 36: Using translations, not hardcoded values.

packages/frontend/src/services/FileUploadService.ts
```diff
@@ -0,0 +1,15 @@
+import http from 'axios';
+
+const upload = (nodeId: string, file: File): Promise<any> => {
+  let formData = new FormData();
+
+  formData.append('file', file);
+
+  return http.post(`http://localhost:3004/api/alfresco/file/hallintaraportit`, formData);
```
mehtis, line 8: Pass on the category class. Hardcoded now.

packages/server/lambdas/alfresco/delete-file.ts
```diff
@@ -0,0 +1,62 @@
...
+import { getUser, validateReadUser, validateWriteUser } from '../../utils/userService';
+import { DatabaseClient } from '../database/client';
+import { deleteFileRequestBuilder } from './fileRequestBuilder';
+
+const database = await DatabaseClient.build();
+
+let fileEndpointsCache: Array<CategoryDataBase> = [];
+
+/**
+ * Update custom content for page. Example request: /api/alfresco/file/linjakaaviot/test.pdf
```
mehtis, line 15: Should probably say `Remove custom content for page.` or something to that effect.

packages/server/lambdas/alfresco/delete-file.ts
```diff
@@ -0,0 +1,62 @@
...
+    // TODO: Alfresco functionality. See search.ts for help
+    const request = deleteFileRequestBuilder({ fileName: fileName, category: category });
+
+    return {
+      statusCode: 200,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      // TODO: Return content
+      body: JSON.stringify('Appropriate information about file as JSON'),
```
mehtis, line 56: Probably should do something about this TODO. What about the other TODO nine rows above?

packages/server/lambdas/alfresco/fileRequestBuilder/__tests__/index.test.ts
```diff
@@ -0,0 +1,44 @@
...
+        isBase64Encoded: true,
+        headers: {
+          accept: 'multipart/form-data',
+          'content-length': '360',
+          'content-type': 'multipart/form-data; boundary=X-INSOMNIA-BOUNDARY',
+          host: 'localhost:3004',
+          'oam-remote-user': 'LX348422',
+          'user-agent': 'insomnia/2022.7.2',
+          'x-amzn-trace-id': 'Root=1-63d132c6-4ffc016c2cc7a55547c684dd',
+          'x-api-key': '5sr7ArpBUa7vuHnuCpTZr5I5XIGV9eO78qpp9Tqd',
```
mehtis, line 24: Should this be here?

packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
```diff
@@ -0,0 +1,42 @@
...
+    const parsedForm = (await parseForm(event)) as ParsedFormDataOptions;
+    const options = {
+      method: 'POST',
+      body: createForm(parsedForm),
+      headers: headers,
+    } as unknown as RequestInit;
+    return options;
+  }
+  public deleteRequestBuilder(requestParameters: FileDeleteRequestBody): FileDeleteRequest {
+    throw new Error('Method not implemented.');
```
mehtis, line 40: Delete missing. Now I realise that deletion isn't promised just yet and is to be implemented. 😅

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -0,0 +1,76 @@
...
+import { fileRequestBuilder } from './fileRequestBuilder';
+import fetch from 'node-fetch';
+import { RequestInit } from 'node-fetch';
+
+const database = await DatabaseClient.build();
+
+const fileEndpointsCache: Array<CategoryDataBase> = [];
+
+const postFile = async (options: RequestInit, nodeId: string) => {
+  const url = `https://api.testivaylapilvi.fi/alfresco/api/-default-/public/alfresco/versions/1/nodes/${nodeId}/children`;
```
mehtis, line 18: From ParameterStore, not hardcoded.

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -0,0 +1,76 @@
...
+    const res = await fetch(url, options);
+    const result = await res.text();
+    return result;
+  } catch (err) {
+    console.error('error:' + err);
+  }
+};
+
+/**
+ * Update custom content for page. Example request: /api/alfresco/file/linjakaaviot
```
mehtis, line 29:
```suggestion
 * Upload custom content for page. Example request: /api/alfresco/file/linjakaaviot
```

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    if (!category || paths.pop() !== 'file') {
+      throw new RataExtraLambdaError('Category missing from path', 400);
+    }
+    if (isEmpty(event.body)) {
+      throw new RataExtraLambdaError('Request body missing', 400);
+    }
+    if (!fileEndpointsCache.length) {
+      const categories = await database.categoryDataBase.findMany();
+      categories.forEach((category) => {
```
mehtis, line 51: If this is how it should be, then this should be added to other lambdas as well. Not in this ticket, but in it's own ticket. I originally tested this with only one value available in the DB, so I definitely could have missed this.

Out of interest: Why do we have to forEach this instead of setting the cache as the return value?

packages/server/lambdas/alfresco/update-file.ts
```diff
@@ -0,0 +1,65 @@
...
+    // TODO: Alfresco functionality. See search.ts for help
+    const request = fileRequestBuilder({ ...body, category: category });
+
+    return {
+      statusCode: 200,
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      // TODO: Return content
+      body: JSON.stringify('Appropriate information about file as JSON'),
```
mehtis, line 59: Presumably updating is also coming later?

packages/server/utils/parser.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface ParsedFormDataOptions {
+  [key: string]: string | Buffer | Readable | FileInfo;
+}
+
+export const parseForm = (event: ALBEvent) => {
+  return new Promise<ParsedFormDataOptions>((resolve, reject) => {
+    const form = {} as ParsedFormDataOptions;
+    const bb = busboy({ headers: event.headers });
+
+    // bb.on('field', (fieldname, val) => {
```
mehtis, line 14: Commented out code.

packages/server/utils/parser.ts
```diff
@@ -0,0 +1,40 @@
+import { ALBEvent } from 'aws-lambda';
+import busboy, { FileInfo } from 'busboy';
+import { Readable } from 'stream';
+
+export interface ParsedFormDataOptions {
+  [key: string]: string | Buffer | Readable | FileInfo;
+}
+
+export const parseForm = (event: ALBEvent) => {
```
mehtis, line 9: Did you get streaming working?

mehtis (changes requested):

packages/frontend/src/pages/Others/ManagementReports/index.tsx
```diff
@@ -1,3 +1,48 @@
...
+      }
+
+      setFile(undefined);
+    });
+  };
+  const [file, setFile] = useState<File>();
+  const [nodeId, setNodeId] = useState<string>('');
+  return (
+    <>
+      <TextField type="file" onChange={selectFile} />
```
NinaDang97, line 28: Should file upload accept only a few file formats (PDF, Excel, Image, PowerPoint, Word, PlainText)? You could find `mimeTypesMapping` in file luceneQueryBuilder.ts

e.g. `<TextField type="file" inputProps={{ accept:"application/pdf" }} onChange={selectFile} />`

packages/frontend/src/services/FileUploadService.ts
NinaDang97: Should we have some sort of error handling (e.g. error notification) and spinner loading also?

packages/server/lambdas/alfresco/fileRequestBuilder/__tests__/index.test.ts
NinaDang97: Yeah, has some sensitive data. Could check if some properties exist instead?

packages/server/lambdas/alfresco/upload-file.ts
saneDG: Good point. We don't need to use loop here 👍🏻

packages/frontend/src/pages/Others/ManagementReports/index.tsx
saneDG: Removed this frontend code since it'll not be used in production anyway. We can discuss about this in the future.

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -0,0 +1,74 @@
...
+ * @returns  {Promise<ALBResult>} JSON stringified object of uploaded file metadata
+ */
+export async function handleRequest(event: ALBEvent): Promise<ALBResult | undefined> {
+  try {
+    const paths = event.path.split('/');
+    const category = paths.pop();
+
+    const user = await getUser(event);
+    validateReadUser(user);
+
```
mehtis, line 43: Logging is missing. Should log user, category and file being uploaded.

lib/rataextra-pipeline.ts
```diff
@@ -40,6 +41,37 @@ export class RataExtraPipelineStack extends Stack {
...
+        'rm -rf asset.*',
+        'zip -r -q -A $ZIP_ARCHIVE *',
+        'aws s3 cp $ZIP_ARCHIVE s3://$S3_PATH',
+        'object_location=${S3_PATH#*/}',
+        'for bucket in $cross_region_replication_buckets; do aws s3 cp $ZIP_ARCHIVE s3://$bucket/$object_location; done',
+      ],
+      rolePolicyStatements: [
+        new PolicyStatement({
+          effect: Effect.ALLOW,
+          resources: ['arn:aws:s3:::*'],
```
mehtis, line 60: Definitely better, but there's still room for improvement later on this. Maybe add a TODO here for when there's need to update the bucket to a removable one that this should also be updated then.

mehtis (changes requested):
Need to add logging, should be good after that.

lib/rataextra-pipeline.ts
saneDG: Adding this TODO

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -62,6 +63,7 @@ export async function handleRequest(event: ALBEvent): Promise<ALBResult | undefi
     const requestOptions = (await fileRequestBuilder(event, headers)) as RequestInit;
 
     const result = await postFile(requestOptions, categoryData.alfrescoFolder);
+    log.info(user, `uploading file ${result?.entry.name} to ${categoryData.alfrescoFolder}`);
```
mehtis, line 66: Split this log into two:  First before the validate user, where you log user and alfrescoFolder. Second here where you log user and file name (folder is also fine to have here). This second should also say "Uploaded", as the upload has already been done at this point.

Just a side note that in alfresco search the log.info is after the validation, which should be fixed. If you can do it here as well (no need to move the debugs), that'd be swell.

mehtis (approved):
Such a small a note that I'll already approve this and you can merge after doing the changes.

mehtis (approved):

PR #154, opened 2023-02-13
Feat/RTENU-16 category file downloads

NinaDang97:
- `REACT_APP_ALFRESCO_API_URL` has same value as `rataextra-alfresco-api-url` in parameter store
- I added REACT_APP_ALFRESCO_API_URL as environment variable to stage Build in CodePipeline (Let's see if it works )
- Only pdf is downloaded in preview mode, other file formats will be downloaded


packages/frontend/src/components/Files/File.tsx
```diff
@@ -42,13 +42,24 @@ export const NodeItem = ({ node, row }: any) => {
...
+      sx={{
+        paddingBottom: '18px',
+        backgroundColor: row % 2 ? Colors.lightgrey : Colors.white,
+        cursor: 'pointer',
+        textDecoration: 'none',
+      }}
+      component="a"
+      target="_blank"
+      rel="noopener noreferrer"
+      href={`${process.env.REACT_APP_ALFRESCO_API_URL}/alfresco/versions/1/nodes/${id}/content?attachment=false`}
```
mehtis, line 54: I've been told that it's better to have process.env-calls in a const at the same level as that NodeTypes in this file (or "higher"), as each process.env call is a comparatively slow process, at least if done a lot. So that's what I'll keep on parroting! 😄

packages/frontend/.env.example
```diff
@@ -1 +1 @@
-REACT_APP_API_URL=http://localhost:3001/api
+REACT_APP_RATAEXTRA_ALFRESCO_API_URL=
```
mehtis, line 1: This problem is now solved locally, but how about when deployed to AWS?

mehtis (changes requested):

packages/frontend/src/components/Files/File.tsx
```diff
@@ -32,6 +32,7 @@ export const NodeItem = ({ node, row }: any) => {
   const { id, name, modifiedAt, content } = entry;
   const contentMimeType = get(content, 'mimeType', '');
   const contentSizeInBytes = get(content, 'sizeInBytes', 0);
+  const { REACT_APP_ALFRESCO_DOWNLOAD_URL } = process.env;
```
mehtis, line 35: Haven't seen process.env used like this. Intriguing! What is the value here if that variable is not in env?

NinaDang97: it's `undefined`

mehtis:
Starting to look good. Once you're happy, just re-request a review and we should be good to go.

lib/rataextra-pipeline.ts
```diff
@@ -11,7 +11,7 @@ import { PolicyStatement, Effect } from 'aws-cdk-lib/aws-iam';
  */
 export class RataExtraPipelineStack extends Stack {
   constructor(scope: Construct) {
-    const config = getPipelineConfig();
+    const config = getPipelineConfig(scope);
```
mehtis, line 14: Presumably this scope can be undefined or something here? What if you call `super(scope)` in the line above and pass this to `getPipelineConfig`? That's how it is dome in rataextra-stack.ts. Give that a try, if that removes the need to handle undefined. If not, then no need to make changes. If it does, great. I'm already giving this approval, but do try that first before merging.

mehtis (approved):

lib/rataextra-pipeline.ts
NinaDang97: I let `scope?: Construct` as optional in `getPipelineConfig()` cause cdk.ts doesn't have a class to pass `scope`

mehtis: Oh, I see! Thanks for the explanation. Feel free to merge this now.

PR #162, opened 2023-02-22
handle api responses

saneDG:
- Display message for user based on response recieved from file upload request

image

image

### DIFF
--- a/packages/frontend/src/assets/locales/fi/common.json
+++ b/packages/frontend/src/assets/locales/fi/common.json
@@ -29,6 +29,7 @@
     "description": "Selitysteksti",
     "more_actions": "Valinnaiset toiminnot",
     "file_not_selected": "Tiedostoa ei valittu",
-    "file_not_uploaded": "Tiedostoa ei ladattu"
+    "file_not_uploaded": "Tiedostoa ei ladattu",
+    "file_uploaded": "Tiedosto ladattu"
   }
 }
--- a/packages/frontend/src/assets/locales/fi/common.json
+++ b/packages/frontend/src/assets/locales/fi/common.json
@@ -29,7 +29,7 @@
     "description": "Selitysteksti",
     "more_actions": "Valinnaiset toiminnot",
     "file_not_selected": "Tiedostoa ei valittu",
-    "file_not_uploaded": "Tiedostoa ei ladattu",
-    "file_uploaded": "Tiedosto ladattu"
+    "file_not_uploaded": "Tiedostoa ei lähetetty",
+    "file_uploaded": "Tiedosto lähetetty"
   }
 }
--- a/packages/frontend/src/components/Files/FileUploadDialog.tsx
+++ b/packages/frontend/src/components/Files/FileUploadDialog.tsx
@@ -12,21 +12,23 @@ import { LocaleLang } from '../../constants/Units';
 import { DateFormat } from '../../constants/Formats';
 import { uploadFile } from '../../services/FileUploadService';
 import { ButtonWrapper } from '../../styles/common';
-import { getLocaleByteUnit } from '../../utils/helpers';
+import { getLocaleByteUnit, getRouterName } from '../../utils/helpers';
 
 import { FileInput } from '../FileInput/FileInput';
 import { FileModal } from '../Modal/FileModal';
 
 import { Colors } from '../../constants/Colors';
 import './styles.css';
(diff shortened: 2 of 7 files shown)


packages/frontend/src/assets/locales/fi/common.json
```diff
@@ -29,6 +29,7 @@
     "description": "Selitysteksti",
     "more_actions": "Valinnaiset toiminnot",
     "file_not_selected": "Tiedostoa ei valittu",
-    "file_not_uploaded": "Tiedostoa ei ladattu"
+    "file_not_uploaded": "Tiedostoa ei ladattu",
```
mehtis, line 32: Should uploading be translated as "lähettäminen"? To avoid confusion. Just a suggestion.

packages/frontend/src/components/Files/FileUploadDialog.tsx
```diff
@@ -44,22 +44,27 @@ export const FileUploadDialog = ({ categoryName, open, onClose }: FileUploadProp
...
-          return result;
-        })
-        .catch((error) => {
-          setIsLoading(false);
-          setFileError(error);
-        });
-    }, 100000);
+      .catch((error) => {
+        setIsLoading(false);
+        setError(true);
```
mehtis, line 61: Should `setSuccess(false)` here?

packages/frontend/src/assets/locales/fi/common.json
saneDG: That's right! Good point 👍🏻

packages/frontend/src/components/Modal/FileModal.tsx
```diff
@@ -1,56 +1,75 @@
...
+                onClick={() => handleClose()}
+              >
+                <Close></Close>
+              </IconButton>
+            </Box>
+            {children}
+          </div>
+        </ModalContentWrapper>
+      </Modal>
+      <Snackbar open={error} onClose={onSnackbarClose}>
```
NinaDang97, line 57: I'm thinking with snackbar, adding `autoHideDuration` would better user experience so user doesn't need to close themselves for such notification. Could you move snackbar to notification folder so we can re-use  the same accross the app?

packages/frontend/src/components/Files/FileUploadDialog.tsx
```diff
@@ -44,22 +44,28 @@ export const FileUploadDialog = ({ categoryName, open, onClose }: FileUploadProp
...
+      name,
+      description,
+      parentNode: categoryName.toLocaleLowerCase(),
+    })
+      .then((result) => {
+        setIsLoading(false);
+        handleClose();
+        setError(false);
+        setSuccess(true);
+        return result;
```
NinaDang97, line 57: When success, we also need to refresh the file list to get new result

packages/frontend/src/components/Files/FileUploadDialog.tsx
```diff
@@ -44,22 +44,28 @@ export const FileUploadDialog = ({ categoryName, open, onClose }: FileUploadProp
...
     setIsLoading(true);
-    setTimeout(async () => {
-      await uploadFile(file, {
-        name,
-        description,
-        parentNode: categoryName.toLocaleLowerCase(),
+    await uploadFile(file, {
+      name,
+      description,
+      parentNode: categoryName.toLocaleLowerCase(),
```
NinaDang97, line 50: should it be `getRouterName` from 'utils/helpers'?

NinaDang97 (changes requested):

packages/frontend/src/components/Modal/FileModal.tsx
saneDG: Added <SnackbarAlert/> component

packages/frontend/src/pages/Others/ManagementReports/index.tsx
```diff
@@ -5,13 +5,19 @@ import { getSubCategoryData } from '../../../utils/helpers';
 
 export const ManagementReports = () => {
   const categoryNames = getSubCategoryData();
+  const handleUpload = (result: any) => {
+    console.log(result);
```
NinaDang97, line 9: @saneDG is it for TODO?

NinaDang97 (commented):
👍

NinaDang97 (approved):

PR #163, opened 2023-02-23
Feat/126 update file api

saneDG:
Add file-update endpoint and lambda function.

Updates file content while keeping same id. File name can be edited in same request with `name` parameter.

`PUT` `/api/alfresco/file/{folder}/{nodeId}`

### DIFF
--- a/packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
+++ b/packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
@@ -5,8 +5,13 @@ import { ALBEvent } from 'aws-lambda';
 import { Blob } from 'buffer';
 import { FileInfo } from 'busboy';
 
-const base64ToBuffer = (base64string: string) => {
+const base64ToString = (base64string: string): string => {
   const buffer = Buffer.from(base64string, 'base64').toString('utf-8').replace(/\r?\n/g, '\r\n');
+  return buffer;
+};
+
+const base64ToBuffer = (base64string: string): Buffer => {
+  const buffer = Buffer.from(base64string, 'base64');
   return buffer;
 };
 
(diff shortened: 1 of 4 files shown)


packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
```diff
@@ -36,6 +41,15 @@ export class AlfrescoFileRequestBuilder {
     } as unknown as RequestInit;
     return options;
   }
+  public async updateRequestBuilder(event: ALBEvent, headers: HeadersInit) {
+    const buffer = base64ToBuffer(event.body as string);
+    const options = {
+      method: 'PUT',
+      body: buffer,
+      headers: headers,
+    } as unknown as RequestInit;
```
mehtis, line 50: Curious about the unknown here. Got anything to share?

mehtis (approved):

packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
saneDG: I don't know reason for this either 😕 Must be unnecessary copypaste mistake. I'll remove this before merge.

saneDG: This is actually fixed in #163

packages/server/lambdas/alfresco/update-file.ts
```diff
@@ -0,0 +1,89 @@
...
+    const res = await fetch(url, options);
+    const result = (await res.json()) as AlfrescoResponse;
+    return result;
+  } catch (err) {
+    console.error('error:' + err);
+  }
+};
+
+/**
+ * Upload custom content for page. Example request: /api/alfresco/file/linjakaaviot
```
NinaDang97, line 38: Should be "Update content"? Could you update this comment section?

NinaDang97 (approved):

PR #164, opened 2023-02-23
Feat/126-update-file-metadata

saneDG:
🔔 This PR is pointing to #163 !

- Add update-metadata endpoint and lambda.

- Edit update-file endpoint -> use `../{nodeId}/content` path parameter for content update and `../{nodeId}/` for updating metadata.

### DIFF
--- a/packages/server/lambdas/alfresco/fileRequestBuilder/index.ts
+++ b/packages/server/lambdas/alfresco/fileRequestBuilder/index.ts
@@ -18,6 +18,14 @@ export const updateFileRequestBuilder = (event: ALBEvent, headers: HeadersInit, 
       return alfrescoRequestBuilder.updateRequestBuilder(event, headers);
   }
 };
+export const updateFileMetadataRequestBuilder = (event: ALBEvent, headers: HeadersInit, store?: FileStore) => {
+  switch (store) {
+    case FileStore.ALFRESCO:
+    default:
+      const alfrescoRequestBuilder = new AlfrescoFileRequestBuilder();
+      return alfrescoRequestBuilder.updateJsonRequestBuilder(event, headers);
+  }
+};
 export const deleteFileRequestBuilder = (request: FileDeleteRequestBody, store?: FileStore): FileDeleteRequest => {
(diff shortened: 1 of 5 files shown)


packages/server/lambdas/alfresco/update-file-metadata.ts
```diff
@@ -0,0 +1,82 @@
...
+    const res = await fetch(url, options);
+    const result = (await res.json()) as AlfrescoResponse;
+    return result;
+  } catch (err) {
+    console.error('error:' + err);
+  }
+};
+
+/**
+ * Upload custom content for page. Example request: /api/alfresco/file/linjakaaviot
```
NinaDang97, line 32: Same here "Update metadata for page"

NinaDang97 (approved):

packages/server/lambdas/alfresco/fileRequestBuilder/index.ts
```diff
@@ -18,6 +18,14 @@ export const updateFileRequestBuilder = (event: ALBEvent, headers: HeadersInit,
       return alfrescoRequestBuilder.updateRequestBuilder(event, headers);
   }
 };
+export const updateFileMetadataRequestBuilder = (event: ALBEvent, headers: HeadersInit, store?: FileStore) => {
+  switch (store) {
+    case FileStore.ALFRESCO:
+    default:
```
NinaDang97, line 24: Wonder why these lines are not listed under `case FileStore.ALFRESCO:` but `default:`?

mehtis: Actually, it's both. With no break or return in a case, the switch won't stop going over the list. So this is kinda like `if (case === FileStore.ALFRESCO || case === default) { /* DO STUFF */}`

mehtis (approved):

packages/server/lambdas/alfresco/update-file-metadata.ts
mehtis: ☝️

PR #165, opened 2023-02-24
Feat/126 delete file

saneDG:
New delete-file endpoint and lambda function

### DIFF
--- a/lib/rataextra-backend.ts
+++ b/lib/rataextra-backend.ts
@@ -324,7 +324,7 @@ export class RataExtraBackendStack extends NestedStack {
       },
       {
         lambda: alfrescoDeleteFile,
-        priority: 134,
+        priority: 140,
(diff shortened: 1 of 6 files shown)


lib/rataextra-backend.ts
```diff
@@ -316,6 +322,13 @@ export class RataExtraBackendStack extends NestedStack {
         httpRequestMethods: ['PUT'],
         targetName: 'alfrescoUpdateFileMetadata',
       },
+      {
+        lambda: alfrescoDeleteFile,
+        priority: 134,
```
mehtis, line 327: I'd recommend setting this to 140, so there's more space to manoeuvre if need be.

mehtis (approved):
Very small note, but otherwise already approved.

NinaDang97 (approved):

PR #173, opened 2023-03-06
BUGFIX: add production values

saneDG:
- Add conditional operations that checks if environment is set as `prod`. Set securityGroupId and vpcId based on that.

### DIFF
--- a/lib/utils.ts
+++ b/lib/utils.ts
@@ -30,11 +30,6 @@ export const isFeatOrLocalStack = (rataExtraEnv: RataExtraEnvironment) =>
 export const isLocalStack = (rataExtraEnv: RataExtraEnvironment) => rataExtraEnv === ENVIRONMENTS.local;
 
 /**
- * Returns true if environment is set as prod
- */
-export const isProductionEnvirEnvironment = (rataExtraEnv: string) => rataExtraEnv === ENVIRONMENTS.prod;
-
-/**
  * Returns whether the stack is one of the two permanent RataExtra stacks
  * - development main stack that corresponds to development main branch in Github
  * - producition stack that corresponds to production branch in Github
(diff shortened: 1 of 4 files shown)


lib/utils.ts
```diff
@@ -29,6 +29,11 @@ export const isFeatOrLocalStack = (rataExtraEnv: RataExtraEnvironment) =>
  */
 export const isLocalStack = (rataExtraEnv: RataExtraEnvironment) => rataExtraEnv === ENVIRONMENTS.local;
 
+/**
+ * Returns true if environment is set as prod
+ */
+export const isProductionEnvirEnvironment = (rataExtraEnv: string) => rataExtraEnv === ENVIRONMENTS.prod;
```
mehtis, line 35: Could `isProductionStack` be used? If not, there's extra `Envir` in the name.

lib/utils.ts
```diff
@@ -37,12 +42,14 @@ export const isLocalStack = (rataExtraEnv: RataExtraEnvironment) => rataExtraEnv
...
-  vpcId: 'vpc-092f1d064d39ca6a1',
-  availabilityZones: ['eu-west-1a', 'eu-west-1b'],
-  privateSubnetIds: ['subnet-05ba766fa5c0f0eb0', 'subnet-019cb289645adae50'],
-});
+export const getVpcAttributes = (rataExtraEnv: RataExtraEnvironment) => {
+  const vpcId = isProductionEnvirEnvironment(rataExtraEnv) ? 'vpc-0f7ce3c168bc01755' : 'vpc-092f1d064d39ca6a1';
+  return {
+    vpcId: vpcId,
+    availabilityZones: ['eu-west-1a', 'eu-west-1b'],
+    privateSubnetIds: ['subnet-05ba766fa5c0f0eb0', 'subnet-019cb289645adae50'],
```
mehtis, line 50: SubnetIds are also account specific.

mehtis (changes requested):

lib/utils.ts
mehtis: Thinking again on this, no need to use `isProductionStack`.

mehtis (approved):

PR #182, opened 2023-03-14
add file uploading to every protected page

saneDG:
- Add (yet unstyled) button to every protected page. Allows admin to add files to current page.

### DIFF
--- a/packages/frontend/src/pages/ProtectedPage/index.tsx
+++ b/packages/frontend/src/pages/ProtectedPage/index.tsx
@@ -7,6 +7,9 @@ import { AppBarContext } from '../../contexts/AppBarContext';
 import { SlateInputField } from '../../components/Editor/SlateInputField';
 import { EditorContext } from '../../contexts/EditorContext';
 import { isSlateValueEmpty } from '../../utils/slateEditorUtil';
+import { FileUploadDialogButton } from '../../components/Files/FileUploadDialogButton';
+import { matchRoutes, useLocation } from 'react-router-dom';
+import { routes } from '../../routes';
 
 type Props = {
   children: React.ReactElement;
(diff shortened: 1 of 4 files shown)


packages/frontend/src/pages/ProtectedPage/index.tsx
```diff
@@ -17,6 +20,16 @@ type Props = {
 export const ProtectedPage = ({ children }: Props) => {
   const { openEdit, openToolbar } = useContext(AppBarContext);
   const { value } = useContext(EditorContext);
+  const location = useLocation();
+
+  const getRouteName = () => {
```
NinaDang97, line 25: Good idea! Could you move to utils so that can be reused elsewhere (e.g. EditorContext)?

packages/frontend/src/routes.tsx
```diff
@@ -92,7 +92,7 @@ const OTHERS_ROUTES = [
   getProtectedRoute(Routes.RAILWAY_MONITORING_SERVICE, <RailwayMonitoringService />),
 ];
 
-const routes: RouteObject[] = [
+export const routes: RouteObject[] = [
```
NinaDang97, line 95: Could you create new arrays named `categoryRoutes` and export that instead? 

```
export const categoryRoutes: RouteObject[] = [
...DIAGRAMS_ROUTES,
.....DIAGRAMS_ROUTES,
  ...OPERATION_ROUTES,
  ...SPECIALTY_STRUCTURES_ROUTES,
  ...SAFETY_EQUIPMENT_ROUTES,
  ...CONTACT_INFORMATION_ROUTES,
  ...OTHERS_ROUTES,
]
```

NinaDang97 (approved):

PR #200, opened 2023-03-23
Feat/RTENU-139 toolbar insert link

NinaDang97:
- With write right, user can insert or remove link. Insert link by filling url in window prompt, remove link by hover mouse on the link text and click 'remove link' icon in the popup. "Uusi linkki" will be added if no texts are chosen for the entered url from window prompt.
- Without edit mode, user can click on link text and get go to new tab with opened url

Screenshot 2023-03-23 at 9 31 26

Screenshot 2023-03-23 at 9 30 38


packages/frontend/src/components/Editor/SlateInputField.tsx
```diff
@@ -32,7 +32,7 @@ export const SlateInputField = () => {
         editor={editor}
         value={slateValue}
         onChange={(value: any) => {
-          const isAstChange = editor.operations.some((op) => 'set_selection' !== op.type);
+          const isAstChange = editor.operations.some((op: any) => 'set_selection' !== op.type);
```
mehtis, line 35: Any (hehe) chance to get rid of this any? And even on the value one line above as well. Not necessary, but would be nice.

mehtis (approved):

saneDG (approved):

PR #204, opened 2023-03-28
remove unused test schemas

saneDG:
remove following schemas:
- Profile
- Post


mehtis (changes requested):
I already had https://github.com/finnishtransportagency/ratatiedot-extranet/pull/176 for removing this stuff + more. What it still needed was the migrations, so that the tables will also be removed from the dbs.

saneDG:
PR already exists https://github.com/finnishtransportagency/ratatiedot-extranet/pull/176

PR #205, opened 2023-03-29
Database models for editable components

saneDG:
- Database models for editable components
- Replaced database population script with `seed.ts` file
  - Updated README with steps to seed local database (`npx prisma db seed`)
- Migration files

🚨 Migration needed in dev & prod databases


packages/server/prisma/seed.ts
```diff
@@ -0,0 +1,78 @@
...
+                  ],
+                },
+              },
+            },
+          },
+        ],
+      },
+    },
+  });
+  console.log({ hallintaraportit });
```
mehtis, line 68: console.log

packages/server/prisma/schema.prisma
```diff
@@ -12,16 +12,42 @@ datasource db {
...
+  categoryId       String
+  node             Node?
+  card             Card?
+}
+
+model Node {
+  id                  String             @id @default(uuid())
+  title               String
+  type                String
+  nodeId              String
```
NinaDang97, line 42: What's the difference between Node's id and nodeId?

mehtis (changes requested):
Changes discussed in dms

packages/server/prisma/schema.prisma
saneDG: Idea is that `nodeId` represents Alfresco nodeId and `id` is just id for this table.

NinaDang97: alfrescoNodeId should be best 👍

saneDG: Let's go with `alfrescoNodeId`!

No, CategoryDataContents `baseId` is used [as foreign key to create relation](https://www.prisma.io/docs/concepts/components/prisma-schema/relations) with `CategoryDataBase`. alfrescoNodeId is just value of alfresco  node.

mehtis (approved):
Test this out in dev first before deploying to prod

PR #209, opened 2023-04-04
Feat/RTENU-191 toolbar unit test

NinaDang97:
- Move `getCategoryRouteName` away from helper file as Jest failed to run
- Unit test for SlateToolbar and SlateInputField
- Cypress test for SlateToolbar and SlateInputField

### DIFF
--- a/packages/frontend/src/components/Editor/SlateToolbar.tsx
+++ b/packages/frontend/src/components/Editor/SlateToolbar.tsx
@@ -38,9 +38,11 @@ type BlockButtonProps = {
 };
 
 const MarkButton = ({ editor, format, icon }: MarkButtonProps) => {
+  const { t } = useTranslation(['common']);
+
   return (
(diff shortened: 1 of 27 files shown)


packages/frontend/src/components/Editor/SlateToolbar.tsx
```diff
@@ -85,7 +87,7 @@ export const SlateToolbar = () => {
 
   return (
     <Slate editor={editor} value={value}>
-      <ToolbarPaperWrapper elevation={2}>
+      <ToolbarPaperWrapper elevation={2} aria-label="toolbar">
```
mehtis, line 90: These aria-labels should be translated texts. Maybe you can as @noravux for help?

mehtis (changes requested):

mehtis (approved):

noravux (approved):

PR #210, opened 2023-04-06
Feat/rtenu 191 translate aria labels

noravux:
This PR includes translation support & translations for all current aria-labels.
* Please check the quality of the translations and if they match the context and give suggestions if any. 
* Please also run the Cypress test for SlateToolbar and see if it works correctly as I can't yet run it.


mehtis:
This should be targeted at feat/RTENU-191-toolbar-unit-test

PR #222, opened 2023-05-02
Fix/rtenu 200 display current location

noravux:


packages/frontend/src/components/NavBar/MenuList.tsx
```diff
@@ -18,30 +20,84 @@ export const MenuList = () => {
...
                     key={child.key}
                     href={child.to ? child.to : ''}
+                    selected={child.to === pathname}
                   >
-                    <ListItemText primary={child.primary} />
+                    <ListItemText
+                      primary={child.primary}
+                      sx={{
+                        '.MuiListItemText-primary': {
+                          fontWeight: selected ? 'bold' : 'normal',
```
NinaDang97, line 97: Default font-family "Exo2-Regular" doesn't work with font-weight (I don't know why). To set it bold, you could instead use `fontFamily: 'Exo2-Bold' or 'Exo2-ExtraBold'`

packages/frontend/src/components/NavBar/MenuList.tsx
```diff
@@ -18,30 +20,84 @@ export const MenuList = () => {
...
+                }}
+              >
                 <ListItemIcon>{icon}</ListItemIcon>
-                <ListItemText primary={primary} sx={{ opacity: openDrawer ? 1 : 0 }} />
+                <ListItemText
+                  primary={primary}
+                  sx={{
+                    opacity: openDrawer ? 1 : 0,
+                    '.MuiListItemText-primary': {
+                      fontWeight: selected ? 'bold' : 'normal',
```
NinaDang97, line 64: Same as commented below, you could try using `fontFamily: 'Exo2-Bold' or 'Exo2-ExtraBold'` for bold text

NinaDang97 (changes requested):

NinaDang97:
Otherwise LGTM 👍, I'll approve after this fix

saneDG (approved):

packages/frontend/src/components/NavBar/MenuList.tsx
noravux: Weird, it works for me 🤔 If I set it to either of those two, then it won't work. What browser are you using where the bold font doesn't work with just the `bold` setting?

NinaDang97: I'm using chrome 112.0.5615.137

NinaDang97 (approved):

packages/frontend/src/components/NavBar/MenuList.tsx
noravux: Ok it works now. But yeah, weird that the `fontWeight: bold` works as it should on Firefox but not on Chrome. I used the fontFamily option to go around this for now 👍

PR #223, opened 2023-05-02
feat/RTENU-192-filelist-endpoint

saneDG:
- add endpoint for listing nodes (files or folders) by ID

`GET api/alfresco/{ID}?type=folder/file`

- refactor getComponents lambda
  - change path from /alfresco to /database
  - change priority number to be more accurate and not in scope of /alfresco reserved numbers

### DIFF
--- a/lib/rataextra-backend.ts
+++ b/lib/rataextra-backend.ts
@@ -261,6 +261,12 @@ export class RataExtraBackendStack extends NestedStack {
       relativePath: '../packages/server/lambdas/alfresco/list-components.ts',
     });
 
+    const getNodesById = this.createNodejsLambda({
+      ...prismaAlfrescoCombinedParameters,
+      name: 'get-nodes',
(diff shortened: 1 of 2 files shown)


lib/rataextra-backend.ts
```diff
@@ -382,6 +388,13 @@ export class RataExtraBackendStack extends NestedStack {
         httpRequestMethods: ['GET'],
         targetName: 'checkUserRightOnPageContents',
       },
+      {
+        lambda: getComponents,
```
mehtis, line 392: Presumably this wasn't used by frontend yet?

mehtis (approved):

NinaDang97 (approved):

lib/rataextra-backend.ts
saneDG: Correct. Should not have any issues with this.

mehtis: So my approval stands.

PR #229, opened 2023-05-09
Feat/RTENU-214 map component

NinaDang97:
Manually use https://mapshaper.org/ to convert Shapefile (public/ratatiedot_kunnossapitoalueet_wgs84 folder) to GeoJSON.  Run command `-proj wgs84` to convert again from EPSG:3857 to EPSG:4326 CRS that is compatible for Leaflet to use (public/ratatiedot_kunnossapitoalueet_wgs84.json). 

Only Linjakaaviot page has map view now.
Screenshot 2023-05-09 at 15 54 01

### DIFF
--- a/packages/frontend/src/components/Map/PolylineFinlandMap.tsx
+++ b/packages/frontend/src/components/Map/PolylineFinlandMap.tsx
@@ -11,7 +11,9 @@ export const PolylineFinlandMap = () => {
     const initializeMap = async () => {
       try {
         if (!mapRef.current) {
-          const map = L.map('map').setView([65, 26], 5);
+          const map = L.map('map', {
+            renderer: L.canvas({ tolerance: 6 }),
+          }).setView([65, 26], 5);
 
           L.tileLayer('https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png', {
             attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors',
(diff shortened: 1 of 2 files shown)


packages/frontend/src/utils/mapUtil.ts
```diff
@@ -0,0 +1,59 @@
...
+    } else {
+      throw new Error('Invalid content type: ' + contentType);
+    }
+  } catch (error) {
+    console.log('Error loading GeoJSON:', error);
+    throw error;
+  }
+};
+
+export const getPolylineColor = (kpalue: number) => {
```
saneDG, line 21: I would keep param and variable names in english if possible.

```suggestion
export const getPolylineColor = (maintenanceArea: number) => {
```

saneDG, line 21: I would keep param and variable names in english if possible.

```suggestion
export const getPolylineColor = (maintenanceArea: number) => {
```

packages/frontend/src/components/Map/PolylineFinlandMap.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { loadGeoJson, polygonStyle } from '../../utils/mapUtil';
+
+export const PolylineFinlandMap = () => {
+  const mapRef = useRef<L.Map | null>(null);
+
+  useEffect(() => {
+    const initializeMap = async () => {
+      try {
+        if (!mapRef.current) {
+          const map = L.map('map').setView([65, 26], 5);
```
saneDG, line 14: This adds "padding" to geoJSON lines so they would be easier to click. Might be out of scope but this would be easy to implement.

```suggestion
          const map = L.map('map', {
            renderer: L.canvas({tolerance: 6}),
          }).setView([65, 26], 5);
```

saneDG (approved):

noravux (approved):

PR #234, opened 2023-05-12
Feat/RTENU-216 map area list alfresco folder

NinaDang97:


packages/frontend/src/components/Map/types.ts
```diff
@@ -0,0 +1,9 @@
+import { Feature } from 'geojson';
+
+export interface FinlandMapFeature extends Feature {
+  properties: {
+    id: number;
+    kpalue: number;
+    tilirataos: number;
```
mehtis, line 7: As previously, these should be translated. Though I don't quite know what the translations would be. Any idea @saneDG? 🤔

saneDG: These are tricky.. 
kpalue -> maintenanceArea
tilirataos(tilirataosa??) -> ???

NinaDang97: This is a bit tricky, Finnish keys represent original Json data

mehtis: Right. In that case I'd say this is fine then.

PR #236, opened 2023-05-15
Feat/RTENU-215 static area list

NinaDang97:
Screenshot 2023-05-15 at 16 26 14


packages/frontend/src/components/Folders/StaticAreaFolderList.tsx
noravux, line 1: Should these have translation support?

PR #240, opened 2023-05-22
Hotfix/RTENU-224 category subfolders listing

NinaDang97:
- Replace `subcategory` folder name with `nestedFolderId` or `childFolderName`  to avoid possible duplicated names and is still based on main category name
- Validate and throw error if receive invalid parameter names
- Endpoints:
1. Case 1: Get the list of files and folders embedded to category page.
- Example: /api/alfresco/files?category=linjakaaviot
2. Case 2: Get the list of files and folders embedded to any folder that is a descendant of category page.
- Example: /api/alfresco/files?category=linjakaaviot&nestedFolderId=123
3. Case 3: Get the list of files and folders embedded to direct child folder of category page
- Example: /api/alfresco/files?category=linjakaaviot&childFolderName=vuosi_2023

Screenshot 2023-05-22 at 17 39 54

### DIFF
--- a/packages/server/lambdas/alfresco/list-files.ts
+++ b/packages/server/lambdas/alfresco/list-files.ts
@@ -48,7 +48,7 @@ const searchByTermWithParent = async (uid: string, alfrescoParent: string, page:
     const alfrescoSearchAPIUrl = `${getAlfrescoUrlBase()}/search/versions/1/search`;
     const options = await getAlfrescoOptions(uid, { 'Content-Type': 'application/json;charset=UTF-8' });
     log.info(alfrescoSearchAPIUrl, 'alfrescoSearchAPIUrl');
-    log.info(`bodyRequest ${bodyRequest} is stringified`);
+    log.info(bodyRequest, 'bodyRequest');
     const response = await axios.post(`${alfrescoSearchAPIUrl}`, bodyRequest, options);
     return response.data;
   } catch (err) {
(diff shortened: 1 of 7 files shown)


packages/server/lambdas/alfresco/list-files.ts
```diff
@@ -62,20 +47,45 @@ const searchByTermWithParent = async (
...
+    throw error;
+  }
+};
+
+const isFolderInCategory = async (folderPath: string, category: string) => {
+  // Split the path into its components
+  const pathComponents = folderPath.split('/');
+
+  // Check if the parent folder name is among the path components
+  return pathComponents.includes(category);
```
mehtis, line 80: Can this be specified to a certain index? Because otherwise, if we later on add some hidden archive folder, which could have something like `/archive/linjakaaviot/subfolder`, then this check would return true if you're looking for the subfolder and claim that it's in the "normal" linjakaaviot. Bit of a far fetch, as the user would need to know the id of the subfolder, but it's a possibility.

packages/server/lambdas/alfresco/list-files.ts
```diff
@@ -104,31 +113,46 @@ export async function handleRequest(event: ALBEvent): Promise<ALBResult> {
...
+      const foundFolder = await getFolder(user.uid, folderId);
+      const folderPath = get(foundFolder, 'entry.path.name', '');
+      const isFolderDescendantOfCategory = await isFolderInCategory(folderPath, category);
+      if (isFolderDescendantOfCategory) {
+        data = await searchByTermWithParent(user.uid, folderId, page, language);
+      }
     }
 
+    const responseBody = {
+      hasClassifiedContent: endpoint?.hasClassifiedContent || true,
```
mehtis, line 134: I'm not up-to-date with the classified content stuff, but can fallback `true` here lead to many false positives?

NinaDang97: fallback `true` would just prevent user from manipulating folders/files in the given category context. Removing "true" would probably be better as it was original and would be null otherwise if fallback

mehtis:
My concerns were dealt with, but I'll let someone else also check and approve.

saneDG (approved):

PR #242, opened 2023-05-23
Fix/rtenu 182 error translations

noravux:


packages/frontend/src/utils/errorUtil.ts
```diff
@@ -0,0 +1,19 @@
+import { AxiosError } from 'axios';
+import { Errors } from '../constants/Errors';
+import { t } from 'i18next';
+
+export const getErrorMessage = (error: any): string => {
+  if (error.name === 'AxiosError') {
```
NinaDang97, line 6: or `if(error instanceof AxiosError)`

PR #244, opened 2023-05-24
Hotfix/RTENU-204 filter name tags

NinaDang97:
- Add new texts to landing page + search page when no result is found
- Fix tag's names

### DIFF
--- a/packages/frontend/src/assets/locales/fi/search.json
+++ b/packages/frontend/src/assets/locales/fi/search.json
@@ -14,7 +14,7 @@
   "latest_first": "Uusin ensin",
   "last_searches": "VIIMEISET HAUT",
   "show_results": " Näytetään {{files}}/{{total}} tulosta",
-  "zero_result": "Hakusanalla \"{{query}}\" ei löytynyt tuloksia. Tarkista oikeinkirjoitus, kokeile toisella hakusanalla tai muokkaa valitsemiasi suodattimia. ",
+  "zero_result": "Hakusanalla \"{{query}}\" ei löytynyt tuloksia. Tarkista oikeinkirjoitus, kokeile toisella hakusanalla tai muokkaa valitsemiasi suodattimia.",
   "image": "Kuva",
   "action": {
     "update_results": "Päivitä tulokset",
--- a/packages/frontend/src/assets/locales/fi/search.json
+++ b/packages/frontend/src/assets/locales/fi/search.json
@@ -17,7 +17,7 @@
   "zero_result": "Hakusanalla \"{{query}}\" ei löytynyt tuloksia. Tarkista oikeinkirjoitus, kokeile toisella hakusanalla tai muokkaa valitsemiasi suodattimia.",
   "image": "Kuva",
   "format": "Muoto",
-  "material_class": "Aineistoluokka",
+  "category": "Aineistoluokka",
   "action": {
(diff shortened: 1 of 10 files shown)


packages/frontend/src/assets/locales/fi/search.json
```diff
@@ -14,6 +14,8 @@
   "latest_first": "Uusin ensin",
   "last_searches": "VIIMEISET HAUT",
   "show_results": " Näytetään {{files}}/{{total}} tulosta",
+  "zero_result": "Hakusanalla \"{{query}}\" ei löytynyt tuloksia. Tarkista oikeinkirjoitus, kokeile toisella hakusanalla tai muokkaa valitsemiasi suodattimia. ",
```
mehtis, line 17: Is that space at the end on purpose?

packages/frontend/src/components/Tags/index.tsx
```diff
@@ -1,29 +1,52 @@
...
-  if (fromYear) {
-    yearRange += fromYear;
-    if (fromYear < toYear && toYear) {
-      yearRange += `- ${toYear}`;
+
+  const getYearTagName = (): string => {
+    const fromYear = formatYear(years[0]);
+    const toYear = formatYear(years[1]);
+    if (fromYear) {
+      return fromYear < toYear && toYear ? `${fromYear} - ${toYear}` : fromYear;
```
mehtis, line 17: Shouldn't we first check that is there any toYear before comparing it to fromYear?

NinaDang97: In frontend `toYear` calendar is disabled until `fromYear` is selected so it's reasonable to have it check bit afterwards

mehtis: I meant this bit `fromYear < toYear && toYear`

mehtis: Can `fromYear < toYear` be true if there is not `toYear`?

packages/frontend/src/pages/ContactInformation/InterchangeContactInformation/index.tsx
```diff
@@ -1,12 +1,5 @@
...
 
 export const InterchangeContactInformation = () => {
-  const categoryNames = getSubCategoryData();
-
-  return (
-    <ProtectedContainerWrapper>
-      <PageTitleWrapper>{categoryNames.INTERCHANGE_CONTACT_INFORMATION}</PageTitleWrapper>
-    </ProtectedContainerWrapper>
-  );
+  return <ProtectedContainerWrapper></ProtectedContainerWrapper>;
```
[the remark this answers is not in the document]
NinaDang97: It's intentional for future use, some category pages already have separate content (e.g. Linjakaaviot has map component), some don't yet have but we never know :D

packages/frontend/src/components/Tags/index.tsx
NinaDang97: No, indeed `toYear` check is redundant

mehtis: Do note that toYear is converted to 0 if it is null for that comparison. But presumably fromYear can't be less than 0, so this shouldn't be a problem.

noravux (approved):

PR #259, opened 2023-05-30
feat/RTENU-233-search

saneDG:
- Add prop to `FilterSearch` component that triggers search function when user applies filters.

### DIFF
--- a/packages/frontend/src/components/Search/index.tsx
+++ b/packages/frontend/src/components/Search/index.tsx
@@ -41,6 +41,12 @@ export const Search = ({ isDesktop = false }: SearchProps) => {
     }
   };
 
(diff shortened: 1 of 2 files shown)


packages/frontend/src/components/Search/index.tsx
```diff
@@ -41,6 +41,12 @@ export const Search = ({ isDesktop = false }: SearchProps) => {
     }
```
NinaDang97, line 41: `enterSearch` and `search` functions have duplicate code. Could you place `search()` in `enterSearch`?
```
const enterSearch = (event: React.KeyboardEvent) => {
    if (event.code === 'Enter' && query) {
      search();
    }
```

PR #261, opened 2023-05-31
revert broken change

saneDG:
remove function that was supposed to fix search but ended up breaking it.


mehtis (approved):
Approved just based on code. Haven't tested this.

PR #263, opened 2023-05-31
feat/RTENU-228 remove unused s3 bucket

NinaDang97:
- Define `autoDeleteObjects` and `removalPolicy` properties for buckets in CodePipeline. Other frontend buckets in environments other than production already have this policy.

### DIFF
--- a/lib/rataextra-pipeline.ts
+++ b/lib/rataextra-pipeline.ts
@@ -1,4 +1,4 @@
-import { SecretValue, Stack, Stage, StageProps, Tags } from 'aws-cdk-lib';
+import { RemovalPolicy, SecretValue, Stack, Stage, StageProps, Tags } from 'aws-cdk-lib';
 import { CodeBuildStep, CodePipeline, CodePipelineSource, ShellStep } from 'aws-cdk-lib/pipelines';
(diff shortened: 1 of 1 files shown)


lib/rataextra-pipeline.ts
```diff
@@ -27,8 +29,27 @@ export class RataExtraPipelineStack extends Stack {
...
-    const pipeline = new CodePipeline(this, 'Pipeline-RataExtra', {
-      pipelineName: 'pr-rataextra-' + config.stackId,
+    const artifactBucket = new Bucket(this, `s3-rataextra-pipeline-${config.stackId}`, {
+      autoDeleteObjects: true,
+      removalPolicy: RemovalPolicy.DESTROY,
+    });
+
+    const pipeline = new Pipeline(this, 'pipeline', {
+      artifactBucket: artifactBucket,
+      pipelineName: `pr-rataextra-${config.stackId}`,
```
mehtis, line 39: This should be named differently from the existing pipeline, because when it's merged to main, this name will conflict with the name of the existing pipeline. That's why the initial bit was named `cpl`.

noravux (approved):

PR #266, opened 2023-06-01
hotfix/RTENU-241 support multiple word search

NinaDang97:
Support flexible multiple word search


packages/server/lambdas/alfresco/searchQueryBuilder/luceneQueryBuilder.ts
```diff
@@ -85,7 +85,7 @@ export class LuceneQueryBuilder implements SearchQueryBuilder {
     const wildcardTerms = parameter.term
       .toLowerCase()
       .split(' ')
-      .map((word: string) => `*${word}*`);
+      .map((word: string) => `${word}*`);
```
mehtis, line 88: Tests should also be updated, right?

mehtis (approved):
As mentioned, tests should be updated to the latest change. Otherwise LGTM!

PR #267, opened 2023-06-01
fix/RTENU-238 Show pointer if hovering over file or a folder

noravux:


packages/frontend/src/components/Files/CategoryFiles.tsx
```diff
@@ -124,6 +124,7 @@ export const CategoryFiles = ({ childFolderName, nestedFolderId }: TCategoryFile
           row={index}
           node={node}
           isSelected={isSelected(node)}
+          isStatic={node.entry.isFile || node.entry.isFolder}
```
mehtis, line 127: Any risk of `node` or `entry` being undefined?

noravux: Just checked and shouldn't cause an issue here

mehtis (approved):

PR #270, opened 2023-06-05
Feat/rtenu 249 persist drawer open state

noravux:


packages/frontend/src/components/NavBar/DesktopDrawer.tsx
```diff
@@ -16,27 +16,29 @@ import { Link } from 'react-router-dom';
 import { Routes } from '../../constants/Routes';
 
 export const DesktopDrawer = () => {
-  const { openDrawer, toggleDrawer } = useContext(AppBarContext);
+  const { openDesktopDrawer, toggleDesktopDrawer } = useContext(AppBarContext);
   const { resetMenu } = useContext(MenuContext);
 
-  const toggleDesktopDrawer = () => {
-    if (openDrawer) {
+  const toggleDesktopDrawerr = () => {
```
NinaDang97, line 22: It's generally not recommended to use very similar names for different functions due to confusion, alternative names could be "checkAndToggleDesktopDrawer" or "resetMenuAndToggleDrawer"?

NinaDang97 (changes requested):

NinaDang97 (changes requested):
I leave some suggestions in your PR. Otherwise LGTM!

packages/frontend/src/components/NavBar/DesktopDrawer.tsx
noravux: Oops, that's actually my mistake when I was just testing it out. Gonna fix that.

NinaDang97 (approved):

packages/frontend/src/contexts/AppBarContext.tsx
```diff
@@ -23,7 +25,8 @@ export const AppBarContext = React.createContext({
 export type TUserRight = { canRead: boolean; canWrite: boolean };
 
 export const AppBarContextProvider = (props: any) => {
-  const [openDrawer, setOpenDrawer] = useState(false);
+  const [openMiniDrawer, setOpenMiniDrawer] = useState(false);
+  const [openDesktopDrawer, setOpenDesktopDrawer] = useState(localStorage.getItem('desktopDrawerOpen') === 'true');
```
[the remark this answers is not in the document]
noravux: > What are the benefits of `openDesktopDrawer` that is saved to local storage, for example compared to `const [openDesktopDrawer, setOpenDesktopDrawer] = useState(true)`?

It's to persist the state of the drawer (open or close) between switching pages and page refreshes, since the local state gets cleared on every refresh

PR #271, opened 2023-06-05
Fix/pdf upload

saneDG:
- Replace unnecessary string conversion with Buffer
- Instead of assigning data in parse function directly to form, push the data to Buffer array and concat it at the end.
  - This was the issue that caused too large file uploads to fail

🗒️ This PR fixes file upload to support <1MB files. We are still looking what might cause >1MB file upload to fail.

### DIFF
--- a/packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
+++ b/packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
@@ -16,7 +16,9 @@ const base64ToBuffer = (base64string: string): Buffer => {
 };
 
 const bufferToBlob = (buffer: Buffer) => {
+  devLog.debug('buffer: ' + JSON.stringify(buffer));
   const blob = new Blob([buffer]);
+  devLog.debug('blob: ' + JSON.stringify(blob));
   return blob;
 };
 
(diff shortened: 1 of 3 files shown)


packages/server/utils/parser.ts
```diff
@@ -1,26 +1,41 @@
...
+      },
+    });
+    let form = {} as ParsedFormDataOptions;
 
     bb.on('file', (fieldname: string, file: Readable, fileinfo: FileInfo) => {
-      form.fieldname = fieldname;
-      form.file = file;
-      form.fileinfo = fileinfo as FileInfo;
-
+      const temp: FormData = { file: [], fieldname: '' };
```
mehtis, line 26: Should this also have `fileinfo` here?

packages/server/lambdas/alfresco/fileRequestBuilder/alfrescoRequestBuilder.ts
```diff
@@ -30,13 +31,15 @@ const createForm = (requestFormData: ParsedFormDataOptions): FormData => {
 };
 
 export class AlfrescoFileRequestBuilder {
-  public async requestBuilder(event: ALBEvent, headers: HeadersInit) {
+  public async requestBuilder(event: ALBEvent, headers: ALBEventHeaders) {
+    const body = event.body ?? '';
+    let buffer;
     if (event.isBase64Encoded) {
-      event.body = base64ToString(event.body as string);
+      buffer = base64ToBuffer(event.body as string);
```
mehtis, line 38: Don't we already have body as a const here? No need to access from event.body.

mehtis (approved):

PR #273, opened 2023-06-06
Remove redundant try-catch block from multiple functions

noravux:


packages/server/lambdas/alfresco/delete-folder.ts
```diff
@@ -17,12 +18,9 @@ let fileEndpointsCache: Array<CategoryDataBase> = [];
...
   const alfrescoCoreAPIUrl = `${getAlfrescoUrlBase()}/alfresco/versions/1`;
   const url = `${alfrescoCoreAPIUrl}/nodes/${nodeId}`;
-  try {
-    const res = await fetch(url, options);
-    return res;
-  } catch (err) {
-    log.error('error:' + err);
-  }
+  const res = await fetch(url, options);
+  const result = (await res.json()) as AlfrescoResponse;
```
mehtis, line 22: This returned res instead of result, unlike elsewhere. Any idea if this was on purpose? But this definitely needs to be tested.

mehtis (changes requested):
Discussed elsewhere, please check what I asked. Also one comment which needs to be followed up on.

mehtis (changes requested):
Discussed elsewhere, please check what I asked. Also one comment which needs to be followed up on.

mehtis (changes requested):
Discussed elsewhere, please check what I asked. Also one comment which needs to be followed up on.

packages/server/lambdas/alfresco/delete-file.ts
```diff
@@ -17,14 +17,9 @@ let fileEndpointsCache: Array<CategoryDataBase> = [];
...
-  try {
-    const res = await fetch(url, options);
-    const result = (await res.json()) as AlfrescoResponse;
-    return result;
-  } catch (err) {
-    console.error('error:' + err);
-  }
+  const res = await fetch(url, options);
+  const result = (await res.json()) as AlfrescoResponse;
+  return result;
```
NinaDang97, line 22: "The “Unexpected end of JSON input” error is a syntax error that occurs when the JSON you’re trying to parse is incomplete or invalid." 
So it's possible res.json() encounters an empty or incomplete response. Adding extra try-catch and logging may help, or using axios could avoid "Unexpected end of JSON input" error by returning raw response if not valid JSON?

 ```suggestion
  try {
  const res = await fetch(url, options);
  if (res.ok) {
    const text = await res.text();
    if (!text) return;
    const result = JSON.parse(text);
    return result as AlfrescoResponse;
  } else {
     console.error('HTTP error:', res.status, res.statusText);
     throw new Error(`HTTP error: ${res.status} ${res.statusText}`);
  }
} catch (err) {
   console.error('File deleting failed: ', err);
   throw err;
}
```

Comment (mehtis) [reply packages/server/lambdas/alfresco/delete-file.ts:22]: Good point about checking `res.ok` as well! I'd say this try-catch here is still not needed, as we have the call wrapped in a try-catch in the calling function, but throwing an error when res is not ok is something we should add.

File: packages/server/lambdas/alfresco/delete-folder.ts (hunk 1 above)
Comment (NinaDang97) [inline packages/server/lambdas/alfresco/delete-folder.ts:22]: Same error "Unexpected end of JSON input" could happen here too

File: packages/server/lambdas/alfresco/create-folder.ts [hunk 3]
```diff
@@ -19,12 +19,15 @@ let fileEndpointsCache: Array<CategoryDataBase> = [];
...
-    console.log(err);
+  const res = await fetch(url, options);
+  if (res.ok) {
+    const text = await res.text();
+    if (!text) return;
+    const result = JSON.parse(text);
+    return result as AlfrescoResponse;
+  } else {
+    console.error('HTTP error:', res.status, res.statusText);
+    throw new Error(`HTTP error: ${res.status} ${res.statusText}`);
```
NinaDang97, line 30: Is it possible you could move this line to a new helper function (e.g. alfrescoFetch, apiFetch or something like that?) in util and re-use it?

```
export const alfrescoFetch = (url, options) => ... 
```

NinaDang97 (approved):

packages/server/lambdas/alfresco/delete-folder.ts
noravux: Seems like this function is not currently being used to delete folders, we use `delete-file` instead. This is because alfresco API works that way, it doesn't care if its a file or a folder, same API for both. Maybe we should check in some other ticket how many of these unused functions we have and decide whether to use them for clarity?

mehtis: In that case, maybe add a comment that about this current situation 👍

mehtis (approved):
Besides that comment about comment, LGTM 👍 It's definitely an additional benefit (and SonarQube will agree) that we're able to reduce duplicate code.

PR #275, opened 2023-06-08
Feat/RTENU-254 upload file to subfolder

NinaDang97:
- Move button "Lisää tiedosto" next to "Poista valittu tiedosto", user can only add if have write right and in edit mode
- "Lisää tiedosto" is in every folder so user can upload a file anywhere. New file is added to current file list if succeeded

Screenshot 2023-06-08 at 9 30 02

### DIFF
--- a/packages/frontend/src/services/FileUploadService.ts
+++ b/packages/frontend/src/services/FileUploadService.ts
@@ -22,10 +22,9 @@ export const uploadFile = async (file: File, fileData: FileData): Promise<AxiosR
         'content-type': 'multipart/form-data',
       },
     };
-    response = await axios(
-      `/api/alfresco/file/${nestedFolderId ? `${categoryName}/${nestedFolderId}` : categoryName}`,
-      options,
-    );
+    const originalUrl = `/api/alfresco/file/${categoryName}`;
+    const nestedFolderUrl = `/api/alfresco/file/${categoryName}/${nestedFolderId}`;
+    response = await axios(nestedFolderId ? nestedFolderUrl : originalUrl, options);
   }
   return response as any;
 };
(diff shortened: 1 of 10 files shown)


packages/frontend/src/services/FileUploadService.ts
```diff
@@ -21,7 +22,10 @@ export const uploadFile = async (file: File, fileData: FileData): Promise<AxiosR
         'content-type': 'multipart/form-data',
       },
     };
-    response = await axios(`/api/alfresco/file/${parentNode}`, options);
+    response = await axios(
+      `/api/alfresco/file/${nestedFolderId ? `${categoryName}/${nestedFolderId}` : categoryName}`,
```
mehtis, line 26: Maybe this ternary could be set in a const with a descriptive name?

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -64,7 +72,18 @@ export async function handleRequest(event: ALBEvent): Promise<ALBResult | undefi
...
-    const result = await postFile(requestOptions, categoryData.alfrescoFolder);
+    let result;
+    if (!nestedFolderId) {
+      result = await postFile(requestOptions, categoryData.alfrescoFolder);
+    } else {
+      const foundFolder = await getFolder(user.uid, nestedFolderId);
+      const folderPath = get(foundFolder, 'entry.path.name', '');
+      // Check if the nest folder is a descendant of the category
+      const isFolderDescendantOfCategory = await isFolderInCategory(folderPath, category);
+      if (isFolderDescendantOfCategory) {
```
mehtis, line 83: Should this throw an error, if it's not a descendant? Because in that case, this seems to be throwing an error for failing to JSON.stringify the empty result. Throwing something more descriptive would be more useful for possible debugging.

mehtis (changes requested):

mehtis (approved):

PR #279, opened 2023-06-13
Hotfix/RTENU-256 upload larger file size

NinaDang97:
Attempt to fix "Invalid state: chunk ArrayBuffer is zero-length or detached" error, now can upload file size around 1MB
- Use `axios` and `form-data` instead of `node-fetch` and `formdata-node`, no need to convert Buffer to Blob anymore. Axios uses the `form-data` library under the hood when handling "multipart/form-data" in Node.js so it's more compatible.
- Close stream and clean buffers after use

### DIFF
--- a/packages/server/lambdas/alfresco/upload-file.ts
+++ b/packages/server/lambdas/alfresco/upload-file.ts
@@ -19,7 +19,10 @@ const postFile = async (options: RequestInit, nodeId: string): Promise<AlfrescoR
   const alfrescoCoreAPIUrl = `${getAlfrescoUrlBase()}/alfresco/versions/1`;
   const url = `${alfrescoCoreAPIUrl}/nodes/${nodeId}/children`;
   console.log('URL:', url, ', options:', options);
-  return await alfrescoFetch(url, options);
+  return await alfrescoFetch(url, {
+    ...options,
+    highWaterMark: 1024 * 1024, // ~1MB
+  });
 };
 
 /**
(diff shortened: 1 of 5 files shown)


packages/server/utils/parser.ts
```diff
@@ -14,6 +14,10 @@ export const parseForm = (buffer: Buffer | string, headers: ALBEventHeaders) =>
         ...headers,
         'content-type': headers['Content-Type'] || headers['content-type'],
       },
+      limits: {
+        files: 1,
+        // fileSize: 1000000, // bytes = 1MB
```
mehtis, line 19: Yeah, we don't want to limit this here as we're supposed to support up to 50MB.

packages/server/lambdas/alfresco/upload-file.ts
```diff
@@ -1,24 +1,36 @@
...
+import FormData from 'form-data';
+
+import axios from 'axios';
 
 const database = await DatabaseClient.build();
 
 let fileEndpointsCache: Array<CategoryDataBase> = [];
 
-const postFile = async (options: RequestInit, nodeId: string): Promise<AlfrescoResponse | undefined> => {
+export interface AxiosRequestOptions {
```
mehtis, line 20: Later on this could be moved to some (currently nonexistent) types-file or something. No need just yet, though.

mehtis (approved):
LGTM!

PR #287, opened 2023-06-22
Set desktop drawer to be open by default

noravux:


packages/frontend/src/contexts/AppBarContext.tsx
```diff
@@ -26,7 +26,11 @@ export type TUserRight = { canRead: boolean; canWrite: boolean };
 
 export const AppBarContextProvider = (props: any) => {
   const [openMiniDrawer, setOpenMiniDrawer] = useState(false);
-  const [openDesktopDrawer, setOpenDesktopDrawer] = useState(localStorage.getItem('desktopDrawerOpen') === 'true');
+  const [openDesktopDrawer, setOpenDesktopDrawer] = useState(() => {
+    const storedValue = localStorage.getItem('desktopDrawerOpen');
+    return storedValue !== null ? JSON.parse(storedValue) : true;
```
mehtis, line 31: I did not know that `JSON.parse("true")` returns `true`! TIL! Testing in console, there is the chance that if storedValue is 'truee' it'll throw a syntax error, so I'd still return this to the old check just in case. Less crash-prone.

mehtis (changes requested):

mehtis (approved):

PR #288, opened 2023-06-22
Feat/rtenu 244 file dates

noravux:
~~Still waiting for proper texts to be provided and they have to be added before merging.~~
Correct text has now been added

### DIFF
--- a/packages/frontend/src/pages/Landing/index.tsx
+++ b/packages/frontend/src/pages/Landing/index.tsx
@@ -32,7 +32,10 @@ export const Landing = () => {
       </ParagraphWrapper>
       <SubtitleWrapper variant="subtitle2">Tiedostojen päivämäärät</SubtitleWrapper>
       <ParagraphWrapper variant="body1">
-        Mikäli päivämäärän tilalla on viiva, tarkoittaa se, että aineisto on siirretty vanhasta järjestelmästä massana.
+        Järjestelmämuutoksen (Ratatiedon extranet-&gt;Ratatieto) yhteydessä siirrettiin aineisto uuteen palveluun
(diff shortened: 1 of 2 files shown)


packages/frontend/src/pages/Landing/index.tsx
```diff
@@ -25,10 +25,18 @@ export const Landing = () => {
...
       </ParagraphWrapper>
       <SubtitleWrapper variant="subtitle1">Ajankohtaista</SubtitleWrapper>
+      <SubtitleWrapper variant="subtitle2">Tiedostojen tallentaminen</SubtitleWrapper>
       <ParagraphWrapper variant="body1">
         Yli 1 Mt tiedostojen tallentaminen järjestelmään ei välttämättä toimi ja virheen sattuessa pitää tiedosto
         tallentaa suoraan Alfrescoon.
       </ParagraphWrapper>
+      <SubtitleWrapper variant="subtitle2">Tiedostojen päivämäärät</SubtitleWrapper>
+      <ParagraphWrapper variant="body1">
+        Järjestelmämuutoksen (Ratatiedon extranet-&gt;Ratatieto) yhteydessä siirrettiin aineisto uuteen palveluun
```
mehtis, line 35: Doing line changes inside the text is a bit cumbersome. It would be better to do that in CSS. However, not a blocking change here, just something to think about in the future.

mehtis (approved):

packages/frontend/src/pages/Landing/index.tsx
noravux: Yeah, true. I figured this way would suffice for now since we got the ticket about the editing functionality for the landing page which makes this process easier 😄

PR #304, opened 2023-08-04
RTENU 271 title

saneDG:
Modify query to search files with title

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9037,9 +9037,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001416",
-      "resolved": "https://registry.npmjs.org/caniuse-lite/-/caniuse-lite-1.0.30001416.tgz",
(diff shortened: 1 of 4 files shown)


noravux:
Two problems:
Screenshot 2023-08-07 at 10 15 37
When searching with only a part of the title, no results. Seems to work if I search with the first word of the title but other combinations don't seem to work properly.

Screenshot 2023-08-07 at 10 11 43
When you find the file with the title, the search result list doesn't show the title so you can't really be sure if that is the file you were searching for. Show the title also here.

noravux (changes requested):
Two problems:
Screenshot 2023-08-07 at 10 15 37
When searching with only a part of the title, no results. Seems to work if I search with the first word of the title but other combinations don't seem to work properly.

Screenshot 2023-08-07 at 10 11 43
When you find the file with the title, the search result list doesn't show the title so you can't really be sure if that is the file you were searching for. Show the title also here.

noravux (approved):
tested and LGTM 👍

PR #309, opened 2023-08-15
update content

saneDG:
- Remove 1MB file size limit warning from home page
- Update packages
- Install and fix eslint that has not been working after migrating to Vite. CRA included eslint in itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,15 @@
         "typescript": "~3.9.7"
       }
     },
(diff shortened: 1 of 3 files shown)


antti-honkanen (changes requested):
Muutos muuten ok, mutta tuli asiakkaalta pyyntö poistaa toinenkin tiedote.

antti-honkanen (approved):

PR #315, opened 2023-08-22
Add filtering options to file search

saneDG:
- Add options to search with name/title/description/content
- Add checkboxes to frontend
- Construct query based on request body parameters
  - If none are given, use all of the 4 values mentioned above

image

TODO: fix "clear all filters" button


noravux (commented):
Tested and seems to be working alright, one small quirk I noticed:
Screenshot 2023-08-23 at 9 13 07
I searched for "ääkköset" with description search turned on and it found these files. However when I add "?" at the end, it doesn't seem to find the file anymore 🤔

noravux (approved):

PR #321, opened 2023-08-31
Feat/rtenu 303 folder naming

noravux:


packages/frontend/src/components/Files/File.tsx
```diff
@@ -84,6 +84,8 @@ export const NodeItem = ({
     }
   };
 
+  const nodeTitle = isFolder && title ? title : `${name}${title ? ` (${title})` : ''}`;
+
```
saneDG, line 88: Maybe bit confusing 2 ternary operator in same line but nothing too serious.

saneDG (approved):

PR #349, opened 2023-10-30
Fix/rtenu 318 activity stream fix

noravux:


packages/server/prisma/migrations/20231011101116_init/migration.sql
saneDG, line 1: Migration files could be [squashed](https://www.prisma.io/docs/guides/migrate/developing-with-prisma-migrate/squashing-migrations)

saneDG (approved):

PR #352, opened 2023-11-01
add toaster to inform about missing category filter

saneDG:
- Add new translation
- Show toaster if trying to search with only area filter selected.
- Add function for closing filter

image

### DIFF
--- a/packages/frontend/src/components/Search/index.tsx
+++ b/packages/frontend/src/components/Search/index.tsx
@@ -133,6 +133,3 @@ export const Search = ({ isDesktop = false }: SearchProps) => {
     </>
   );
 };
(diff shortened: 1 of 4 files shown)


packages/frontend/src/components/Search/index.tsx
```diff
@@ -126,3 +133,6 @@ export const Search = ({ isDesktop = false }: SearchProps) => {
     </>
   );
 };
+function closeFilter() {
+  throw new Error('Function not implemented.');
```
noravux, line 137: This function declaration is probably by mistake?

noravux (changes requested):

noravux (approved):

PR #355, opened 2023-11-08
Fix logout session check

saneDG:
- Remove `Return` cookie
- ~~Use `AWSELBAuthSessionCookie-0` & `AWSELBAuthSessionCookie-1` to check authenticated session~~.
- Remove mentions about `Return` cookie


mehtis (approved):
Code changes look good, but check that thing I mentioned about the expiration.

saneDG:
Found potential issues with suggested solution related to too long cookie expiration time. Converting to draft.

PR #360, opened 2023-11-21
Add admin permissions

saneDG:
- Check admin rights and return permissions to read/write


packages/server/lambdas/database/check-user-right.ts
```diff
@@ -22,15 +22,26 @@ export async function handleRequest(event: ALBEvent): Promise<ALBResult> {
...
+      };
+    }
+
     const params = event.queryStringParameters;
     // only listed parameters are accepted
     validateQueryParameters(params, ['category']);
     const category = params?.category;
 
-    const user = await getUser(event);
     log.info(user, `Checking user write permission for page ${category}`);
```
noravux, line 45: Isn't user unreachable here since `const user = await getUser(event);` is now inside the trycatch block?

noravux (changes requested):

noravux (approved):

PR #364, opened 2023-11-23
add get-banners function and endpoint

saneDG:
- new endpoint `GET /api/banners`
  - return every published notice with `showAsBanner` value true


noravux (approved):

noravux:
This is basically identical to the `/api/notices` endpoint, is there a reason we couldn't use that for showing the banners as it returns the exact same data?

PR #368, opened 2023-12-05
Feat/rtenu 322 banner

saneDG:
- Add banner "notifications"
  - Get banners from database and show them as list
image


packages/frontend/src/components/Banner/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const [open, setOpen] = useState(true);
+
+  useEffect(() => {
+    getBanners();
+  }, [getBanners]);
+
+  return (
+    <>
+      <div style={{ margin: '20px 32px 0 32px' }}>
+        <IconButton onClick={() => setOpen(!open)} className="banner-alert">
```
noravux, line 22: I don't think users will realize that you can hide the banner with the Info icon, but we can tweak these small things later.

packages/frontend/src/components/Banner/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+          <InfoIcon color="primary" />
+        </IconButton>
+      </div>
+      {banners?.map((banner) => (
```
noravux, line 26: What if users create multiple notices and set them to be shown as banners? Should we allow only a single banner or just a few banners to be active at same time?

noravux (approved):
Left a few comments but otherwise LGTM, we can tweak these things later

PR #384, opened 2024-01-04
RTENU 339 list order

saneDG:
update area list
- use predefined json file instead of requesting fine name from alfresco
  - old method was slow and every area was dependent on its name


packages/frontend/src/pages/Diagrams/LineDiagrams/index.tsx
```diff
@@ -4,7 +4,7 @@ import { ProtectedContainerWrapper } from '../../../styles/common';
 export const LineDiagrams = () => {
   return (
     <ProtectedContainerWrapper>
-      <PolylineFinlandMap />
+      <PolylineFinlandMap categoryId="71d157b0-6c81-4b55-ad99-d7bfbfd2d960" />
```
noravux, line 7: Is this hardcoded ID here on purpose?

noravux (approved):
Left one comment but LGTM 👍

PR #391, opened 2024-01-16
Feat/rtenu 308 replace file

noravux:
Screenshot 2024-01-16 at 11 11 16
Screenshot 2024-01-16 at 11 11 23

### DIFF
--- a/packages/frontend/src/components/Files/FileUploadDialog.tsx
+++ b/packages/frontend/src/components/Files/FileUploadDialog.tsx
@@ -214,8 +214,7 @@ export const FileUploadDialog = ({
                   variant="contained"
                   disabled={isLoading}
                   onClick={() => {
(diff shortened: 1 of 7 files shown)


packages/frontend/src/components/Files/FileUploadDialog.tsx
```diff
@@ -205,7 +213,14 @@ export const FileUploadDialog = ({ categoryName, nestedFolderId, open, onClose,
                   color="primary"
                   variant="contained"
                   disabled={isLoading}
-                  onClick={() => handleFileUpload(file as File)}
+                  onClick={() => {
+                    const fileAlreadyExists = fileExists(name);
```
saneDG, line 217: Why declare constant here? Could we just use function directly?

saneDG (approved):

PR #480, opened 2024-08-06
main -> prod 08.07.24

saneDG:


packages/node-server/package.json
noravux, line 1: Is there are reason for the `jwk-to-pem` downgrade? 1.2.0 was released 9 years ago 😬 Same in the `packages/server/package.json`

noravux (commented):
Left one comment that needs clarification

noravux (approved):

PR #524, opened 2025-07-07
chore: Add monitoring stack

yessergirem:


yessergirem:
Next step is to configure the monitoring to focus on only the most important services.

PR #526, opened 2025-07-15
Add Sentry integration

yessergirem:


henrinie-nc:
Will not be implemented

PR #636, opened 2026-03-16
feat/RTENU-506: remove single balise create feature

andrea-peltokorpi-nc:
- Remove create balise page routing and navigation button
- Adjust bulk upload navigation button
- Strip now unused create mode logic from balise form


henrinie-nc (approved):
lgtm

PR #646, opened 2026-04-02
Improve logging

juho-kettunen-nc:
This helps developers investigate issues, and avoids unnecessary technical details in user-facing messages.

### DIFF
--- a/packages/server/lambdas/database/error/databaseError.ts
+++ b/packages/server/lambdas/database/error/databaseError.ts
@@ -1,5 +1,6 @@
 import { Prisma } from '../../../generated/prisma/client';
 import { RataExtraLambdaError } from '../../../utils/errors';
+import { log } from '../../../utils/logger';
 
 export type PrismaError = Prisma.PrismaClientKnownRequestError | Prisma.PrismaClientUnknownRequestError;
 
@@ -14,5 +15,6 @@ export const handlePrismaError = (error: PrismaError) => {
     throw new RataExtraLambdaError(`Unknown Prisma client error \n${error.message}`, 500);
   }
 
-  throw new RataExtraLambdaError(`Unknown error`, 500);
+  log.error('Unexpected Prisma error in handlePrismaError:', error);
+  throw new RataExtraLambdaError('Unknown error', 500);
 };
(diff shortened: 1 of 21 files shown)


juho-kettunen-nc:
Now that I think of it, I'll instead use logger to log the error messages to logs, instead of potentially sending unnecessary details to user.

packages/server/lambdas/database/error/databaseError.ts
```diff
@@ -14,5 +15,6 @@ export const handlePrismaError = (error: PrismaError) => {
     throw new RataExtraLambdaError(`Unknown Prisma client error \n${error.message}`, 500);
   }
 
-  throw new RataExtraLambdaError(`Unknown error`, 500);
+  log.error('Unexpected Prisma error in handlePrismaError:', error);
```
henrinie-nc: Perhaps worth double checking hte logger, does it take error in like this as second argument, or should it be stringified in the first argument?

henrinie-nc (changes requested):
Could you check the usage of logger, I don't remember what or how it was setup. Does it take multiple arguments like console.log or what?

packages/server/lambdas/database/error/databaseError.ts
juho-kettunen-nc: Good call. The parameter order is indeed different; object should go first, then message

```
# pino.d.ts > pino.BaseLogger
 /**
         * Log at `'error'` level the given msg. If the first argument is an object, all its properties will be included in the JSON line.
         * If more args follows `msg`, these will be used to format `msg` using `util.format`.
         *
         * @typeParam T: the interface of the object being serialized. Default is object.
         * @param obj: object to be serialized
         * @param msg: the log message to write
         * @param ...args: format string values when `msg` is a format string
         */
        error: LogFn;
 ```

juho-kettunen-nc: There's another override for the simple case, with only a single string parameter.

juho-kettunen-nc: Fixed in d4ca9195a49d75a79fc4d1eadb340264d9bbb91d

juho-kettunen-nc:
Requested changes applied @henrinie-nc

henrinie-nc (approved):

henrinie-nc:
@juho-kettunen-nc approved

PR #657, opened 2026-05-26
Feat/RTENU-518: six digit balise files

andrea-peltokorpi-nc:
- Allow six digit balise files through validation
- Treat sixth balise digit as suffix (similar to K-suffix previously)
- Match six digit balise files to balise ID based on first 5 digits (eg 100031.il -> balise ID 10003)

### DIFF
--- a/packages/server/lambdas/balise/__tests__/add-balise.test.ts
+++ b/packages/server/lambdas/balise/__tests__/add-balise.test.ts
@@ -86,7 +86,7 @@ describe('validateUploadedFiles', () => {
       expect(errors).toHaveLength(0);
     });
 
-    it('should reject files with invalid suffix (not K)', () => {
+    it('should reject files with invalid suffix (not K or digit)', () => {
       const files: ParsedFileUpload[] = [
         {
           filename: '12345X.il',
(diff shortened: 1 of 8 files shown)


packages/server/utils/balise/baliseUtils.ts
```diff
@@ -32,31 +32,33 @@ export function isValidBaliseIdRange(baliseId: number): boolean {
...
+ *   "100034.il" → 10003
+ *   "100030.il" → 10003
+ *   "10000X.il" → null (invalid - only K or digit allowed)
  */
 export function parseBaliseIdFromFilename(filename: string): number | null {
-  // Match: digits, optionally followed by K/k, then a dot and extension
-  // This ensures only K suffix is allowed after the ID
-  const match = filename.match(/^(\d+)[Kk]?\.(il|leu|bis)$/i);
+  // Match: exactly 5 digits (the ID), optionally followed by K/k or a single digit, then a dot and extension
+  const match = filename.match(/^(\d{5})[Kk\d]?\.(il|leu|bis)$/i);
```
juho-kettunen-nc, line 49: This regex string would benefit from being a constant, because now it's defined twice in two different functions, and it's difficult to notice differences by eye.

packages/server/lambdas/balise/__tests__/add-balise.test.ts
```diff
@@ -122,6 +122,62 @@ describe('validateUploadedFiles', () => {
...
+    });
+
+    it('should reject digit suffix files when balise ID does not match', () => {
+      const files: ParsedFileUpload[] = [
+        {
+          filename: '100034.il',
+          buffer: Buffer.from('test'),
+        },
+      ];
+      const errors = validateUploadedFiles(files, 10004);
```
juho-kettunen-nc: Maybe this validation ID should be the same as in the input? Meaning, to change `10004` to `100034`. That might drive home the point that the entire `100034` will not be used as an ID.

juho-kettunen-nc (approved):
Approved, with optional suggestions.

PR #658, opened 2026-05-26
Feat/RTENU-522: four digit balises

andrea-peltokorpi-nc:
- Adjust backend validation to allow balises with four digits
- Adjust frontend validation to allow balises starting from 9000
- Rebuild section logic to use prefix (first 1-2 digits of ID) instead of manual range
- Sort all temporary balises to bottom of table (instead of four digit balises at top)
- Add isTemporary to balise model, treat balises with ID starting with 9 as temporary (to simplify sorting)


packages/server/prisma/schema.prisma
```diff
@@ -109,6 +109,7 @@ model Balise {
   lockReason      String?
   deletedAt       DateTime?
   deletedBy       String?
+  isTemporary     Boolean         @default(false)
```
henrinie-nc, line 112: This could be a bit vague naming, would it make sense to have e.g. a `type` column, and by default all would be some "regular" basile, and then this one would have type "temporary" ? What do you think?

packages/frontend/src/modules/balise/pages/BaliseListPage/BaliseListPage.tsx
```diff
@@ -97,9 +98,18 @@ export const BaliseListPage: React.FC = () => {
       } else {
         // Get all selected section ranges
         const selectedSectionDetails = sectionOptions.filter((section) => selectedSections.includes(section.key));
+        const idMins: number[] = [];
+        const idMaxes: number[] = [];
+        for (const s of selectedSectionDetails) {
```
henrinie-nc, line 103: Just a habit, but perhaps shortening the name to `s` brings no benefit - potentially loses readability.

packages/server/utils/balise/baliseUtils.ts
```diff
@@ -24,41 +24,79 @@ export function isValidExtension(filename: string): boolean {
 
 /**
  * Validate balise ID range
- * Valid range: 10000-99999
+ * Valid range: 9000-99999
  */
 export function isValidBaliseIdRange(baliseId: number): boolean {
```
henrinie-nc, line 29: Are these helpers the same and are they duplicated in `packages/frontend/src/modules/balise/utils/baliseValidation.ts` Please check if there are duplicates and leave only one copy

henrinie-nc (approved):
Looks good overall, I would consider the `isTemporary` column naming and data type, would it be easier to understand if it had some enum to choose from that would explain the "balise type"?

Then some helpers seemed like they were duplicated, please check those.

Approving!

packages/server/prisma/schema.prisma
andrea-peltokorpi-nc: This could work too. This was done as a workaround to have a lightweight way to sort balises starting with 9 to the end. I'm not sure if there's a current use case for other variations (for examble K-suffix balises are only files within balises, ie balise 10003 has file 10003K.il - doesn't affect the database entry). However, I don't know whether that would change in the future.

packages/server/utils/balise/baliseUtils.ts
andrea-peltokorpi-nc: Yeah, some of the validation is duplicated in backend and frontend, and it's not really optimal. I believe in the past we discussed creating a shared utils package, but that it could potentially be a bigger task due to the project structure.

This is also noted on the ticket https://extranet.vayla.fi/jira/browse/RTENU-479

henrinie-nc: Agreed, so lets leave it for later

packages/server/prisma/schema.prisma
henrinie-nc: I see, then it is not probably worthwhile to do anything else but keep it as it is

PR #675, opened 2026-08-18
Fix/rtenu 528 node native dependency management

anssi-aijo-nc:


lib/rataextra-pipeline.ts
```diff
@@ -103,14 +108,20 @@ export class RataExtraPipelineStack extends Stack {
           },
         }),
         buildEnvironment: {
-          buildImage: LinuxBuildImage.STANDARD_7_0,
```
henrinie-nc: It seems to me that the tag `node:22` points to 22.23.2. The good would be that it would update the node version, but bad if that version happened to be compromised.
With specific versions we need to keep updating, and it is quite a core across multiple files (one tends to forget one or two)

juho-kettunen-nc (approved):

PR #677, opened 2026-08-21
Feat/rtenu 534 safety folder

anssi-aijo-nc:


packages/frontend/src/types/types.d.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
henrinie-nc: Is there some reason why this was added?
Also does this apply to the whole file or just the next line?

henrinie-nc (approved):
Looks good, please review the eslint-disable in types.d.ts before merging

packages/frontend/src/types/types.d.ts
anssi-aijo-nc: After making changes to the file commits were blocked by eslint in pre-commit hook. I removed the file level disable and moved it to next line disables where needed.

PR #678, opened 2026-08-24
RTENU-526 and 534

anssi-aijo-nc:
Visual separation of archived notes.
Safety folder.


henrinie-nc (changes requested):
please check my comments, once those are cleared we can merge

henrinie-nc (approved):

